PR #6, opened 2020-03-11
Add the ability to filter alerts by namespace.

cyriltovena:
Signed-off-by: Cyril Tovena <cyril.tovena@gmail.com>


cortex-mixin/alerts.libsonnet
```diff
@@ -163,11 +168,11 @@ local windows = [
           {
             alert: 'CortexCacheRequestErrors',
             expr: |||
-              100 * sum(rate(cortex_cache_request_duration_seconds_count{status_code=~"5.."}[1m])) by (namespace, job, method)
```
slim-bean: I think this is an error, removing the range for the rate?

PR #11, opened 2020-03-16
querier.parallelise-shardable-queries lib support

owen-d:
## What
Introduces library support for `querier.parallelise-shardable-queries`. The new jsonnet config variable, `sharded_queries_enabled` (default false) will adjust component functionality to work more ideally for sharding enabled envs.

## Why

This will allow opt in support for sharded query optimizations, enabling incremental rollouts of sharding support for different environments.

### DIFF
--- a/cortex/config.libsonnet
+++ b/cortex/config.libsonnet
@@ -34,8 +34,25 @@
     // concern recent (ingester) data, which isn't sharded. Therefore, we must strike a balance
     // which allows us to process more sharded queries in parallel when requested, but not overload
     // queriers during normal queries.
-    querierConcurrency: if self.sharded_queries_enabled then 16 else 8,
-    querier_ingester_streaming_enabled: $._config.storage_engine != 'tsdb',
+    querier: {
+      replicas: if $._config.queryFrontend.sharded_queries_enabled then 12 else 6,
+      concurrency: if $._config.queryFrontend.sharded_queries_enabled then 16 else 8,
+      ingester_streaming_enabled: $._config.storage_engine != 'tsdb',
+    },
+
+    queryFrontend: {
+      replicas: 2,
+      shard_factor: 16,  // v10 schema shard factor
+      sharded_queries_enabled: false,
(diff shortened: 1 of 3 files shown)


cortex/querier.libsonnet
```diff
@@ -1,6 +1,10 @@
 {
   local container = $.core.v1.container,
 
+  querier_params:: {
```
gouthamve: We should roll this up into _config.

We can basically do something like this here:
```
_config+:: {
  querier:: {
    .... config options.
  },
}
```

And can be used as `$._config.querier.replicas`

cortex/query-frontend.libsonnet
```diff
@@ -1,5 +1,21 @@
 {
   local container = $.core.v1.container,
+  local deployment = $.apps.v1beta1.deployment,
```
gouthamve: These should be moved closer to where they are being used. Its a weird best-practice we use (_don't look at me_) :)

cortex/query-frontend.libsonnet
```diff
@@ -1,5 +1,21 @@
...
+  local service = $.core.v1.service,
+
+  query_frontend_params:: {
```
gouthamve: Same thing about rolling it up into `_config`.

The idea is that anything configurable should be in `_config`

cortex/config.libsonnet
```diff
@@ -122,8 +129,9 @@
       'querier.batch-iterators': true,
 
       // Don't query ingesters for older queries.
-      // Chunks are 6hrs right now.  Add some slack for safety.
-      'querier.query-ingesters-within': '12h',
+      // Chunks are 6hrs right now.  Add some slack for safety although not too much
+      // if sharded queries are enabled because they only shard non ingester queries.
+      'querier.query-ingesters-within': if $._config.sharded_queries_enabled then '6h15m' else '12h',
```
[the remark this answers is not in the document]
owen-d: We can do that and then see if we're being bottlenecked by 12h segments.

PR #12, opened 2020-03-17
Alerts for WAL metrics

codesome:


codesome:
I tried to take care of it with the range vector (but things like `CortexWALCorruption` should be immediately alerted). I was having a hard time figuring out a combination of functions and `for` period. 

I will take a dig again to use the `for` period.

cortex-mixin/alerts.libsonnet
```diff
@@ -226,6 +226,81 @@ local windows = [
...
+              message: |||
+                {{ $labels.namespace }}/{{ $labels.instance }} failed to delete checkpoint.
+              |||,
+            },
+          },
+          {
+            // 2 or more failed checkpoint deletion in 1h means something is wrong.
+            alert: 'CortexCheckpointDeletionFailed',
+            expr: |||
+              increase(cortex_ingester_checkpoint_deletions_failed_total[1h]) > 1
```
gouthamve, line 293: Can we make this `2h`?

cortex-mixin/alerts.libsonnet
```diff
@@ -226,6 +226,81 @@ local windows = [
               |||,
             },
           },
+          {
+            // Alert immediately if WAL is corrupt.
+            alert: 'CortexWALCorruption',
+            expr: |||
+              rate(cortex_ingester_wal_corruptions_total[5m]) > 0
```
gouthamve, line 233: Why are we using rate here and increase everywhere else?

codesome: Does not matter actually, I will make it increase to be consistent

PR #21, opened 2020-03-31
Remove querier podAntiAffinity

Duologic:


tomwilkie:
Why?  The anti affinity exists for a reason - there is very little point in running multiple queries on the same node when we don't have CPU limits.

tomwilkie:
Why?  The anti affinity exists for a reason - there is very little point in running multiple queries on the same node when we don't have CPU limits.

Duologic:
Are you suggesting we implement CPU limits to cover for this?
What do you think about relaxing the podAntiAffinity from `requiredDuringSchedulingIgnoredDuringExecution` to `preferredDuringSchedulingIgnoredDuringExecution`?

tomwilkie:
> Are you suggesting we implement CPU limits to cover for this?

I don't know what "this" is - what are you trying to acheive?

As it stands, I don't believe we can put CPU limits in place, as there is a well know kernel/kubernetes bug that causes them to overly throttle containers, resulting in terrible latency and utilisation.

https://medium.com/omio-engineering/cpu-limits-and-aggressive-throttling-in-kubernetes-c5b20bd8a718

> What do you think about relaxing the podAntiAffinity from requiredDuringSchedulingIgnoredDuringExecution to preferredDuringSchedulingIgnoredDuringExecution?

AFAIU that will allow multi queriers to run on the same machine, and I don't think thats a good idea.

Duologic:
Can you elaborate on why that is not a good idea?

Duologic:
The reason for proposed change is that querier pods allocate quite a lot of resources and moving them to bigger nodes (running more querier pods on fewer nodes) results in a cheaper infrastructure. 
During the last upgrade, the scheduler wasn't very effective in spreading out the pods and required us to increase the size of the node pool to match the querier allocation requests, at that point @rfratto questioned the AntiAffinity rule for querier too.
The resource allocation (requests) should make sure the querier has enough resources at its disposal, even if we run multiple querier pods per node. As the CPU/Mem utilization is usually fairly low, they can still benefit from it during bursts of requests.

tomwilkie:
> The reason for proposed change is that querier pods allocate quite a lot of resources and moving them to bigger nodes (running more querier pods on fewer nodes) results in a cheaper infrastructure.

The seems odd.  Its my understanding that Google effectively charges per core and per GB of ram, so two small machines cost the same as one machine twice the size.  Hence I'm not sure how running two machines with two queriers is cheaper than running one machine with two queriers, assuming you give queriers the same amount of resources.

> During the last upgrade, the scheduler wasn't very effective in spreading out the pods and required us to increase the size of the node pool to match the querier allocation requests, at that point @rfratto questioned the AntiAffinity rule for querier too.

It sounds like we just need to scale down the number of queriers we run.

> The resource allocation (requests) should make sure the querier has enough resources at its disposal, even if we run multiple querier pods per node. As the CPU/Mem utilization is usually fairly low, they can still benefit from it during bursts of requests.

Queriers are inherently very spiky in their resource usage.  Resource requests are only used in scheduling, they have little to no impact on the actual resources available to a container when it lands on a machine.

woodsaj:
lets move this to https://github.com/grafana/deployment_tools/issues/2977

PR #22, opened 2020-04-01
Add PodDisruptionBudget for ingester

codesome:
With `maxUnavailable` as 1.

### DIFF
--- a/cortex/ingester.libsonnet
+++ b/cortex/ingester.libsonnet
@@ -65,6 +65,7 @@
     deployment.new('ingester', 3, [$.ingester_container], $.ingester_deployment_labels) +
     $.util.antiAffinity +
     $.util.configVolumeMount('overrides', '/etc/cortex') +
+    deployment.mixin.metadata.withLabels({ name: 'ingester' }) +
     deployment.mixin.spec.withMinReadySeconds(60) +
     deployment.mixin.spec.strategy.rollingUpdate.withMaxSurge(0) +
     deployment.mixin.spec.strategy.rollingUpdate.withMaxUnavailable(1) +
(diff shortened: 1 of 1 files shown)


cortex/ingester.libsonnet
```diff
@@ -76,4 +76,13 @@
...
   ingester_service:
     $.util.serviceFor($.ingester_deployment, $.ingester_service_ignored_labels),
+
+  local podDisruptionBudget = $.policy.v1beta1.podDisruptionBudget,
+
+  ingester_pdb:
+    podDisruptionBudget.new() +
+    podDisruptionBudget.mixin.metadata.withName('ingester-pdb') +
+    podDisruptionBudget.mixin.metadata.withLabels({ name: 'ingester-pdb' }) +
+    podDisruptionBudget.mixin.spec.selector.withMatchLabels({ name: 'ingester' }) +
```
joe-elliott, line 86: Is there anyway to use a local variable?  Or perhaps a field on the deployment itself to coordinate this value.  If the "name: ingester" label should ever drift on the ingester pods then this would no longer work.

codesome: Made it a part of `_config`, looks good?

joe-elliott: Apologies for the nit, but I don't think we want this configurable.  just a 
```
local ingesterName = 'ingester'
```
Also something like this is possible too which would require no new vars:
```
podDisruptionBudget.mixin.spec.selector.withMatchLabels({ name: $.ingester_deployment.metadata.labels.name })
```

codesome: > I don't think we want this configurable

Okay. I will try your suggestion with no new variable.

PR #24, opened 2020-04-02
Add playbooks for alerts.

cstyan:
Moving what I can to the public mixin in terms of playbooks for alerts.

### DIFF
--- a/cortex-mixin/docs/playbooks.md
+++ b/cortex-mixin/docs/playbooks.md
@@ -1,0 +1,67 @@
+# Playbooks
+This document contains playbooks, or at least a checklist of what to look for, for alerts in the cortex-mixin.
+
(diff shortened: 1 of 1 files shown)


cortex-mixin/docs/playbooks.md
```diff
@@ -0,0 +1,70 @@
...
+@todo
+
+## CortexQuerierCapacityFull
+@todo
+
+## CortexFrontendQueriesStuck
+@todo
+
+## CortexIngesterRestarts
+@todo
```
codesome, line 64: Repeated

cstyan: good catch, that was just copy/pasted from out internal playbook :+1:

PR #26, opened 2020-04-03
Add an alert for when querier index cache churn rate is too high.

cstyan:
It's been a while since I did anything interesting with queries/label substitution and alerts, so let me know what we could change here to make sure this alert is useful. I based the `entries in cache < 10m` bit based on our env with the lowest value for `entries in cache / churn rate / 60s` being an average of `entry in cache for 12+ minutes`, but for other environments we have anything from 30-60 minutes.

### DIFF
--- a/cortex-mixin/alerts.libsonnet
+++ b/cortex-mixin/alerts.libsonnet
@@ -226,6 +226,22 @@ local windows = [
               |||,
             },
           },
+          {
+            alert: 'CortexQuerierIndexCacheChurnRate',
+            expr: |||
+              (sum(querier_cache_added_new_total{cache="store.index-cache-read.fifocache", job=~".*/querier"}) by (cluster, namespace, job)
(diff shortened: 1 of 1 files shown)


cstyan:
bump @gouthamve

cortex-mixin/alerts.libsonnet
```diff
@@ -226,6 +226,22 @@ local windows = [
               |||,
             },
           },
+          {
+            alert: 'CortexQuerierIndexCacheChurnRate',
```
gouthamve, line 230: The metrics here are for the in-mem fifo-cache imo. We kinda don't care about the fifocache churn rate, but rather about the memcache churn rates. 

The query should be `sum without (job) (memcached_current_items) / sum without (job) (rate(memcached_items_evicted_total[5m])) < 3600`.

It shouldn't be for just index cache, but rather all caches.

gouthamve (changes requested):

PR #37, opened 2020-04-08
Graph ops/sec for "iter" operation

pstibrany:
"iter" doesn't have timing information, so graph ops/sec instead

### DIFF
--- a/cortex-mixin/dashboards.libsonnet
+++ b/cortex-mixin/dashboards.libsonnet
@@ -50,8 +50,9 @@ local g = (import 'grafana-builder/grafana.libsonnet') + {
       g.latencyPanel(operationDuration, '{cluster=~"$cluster", operation="objectsize"}'),
     )
     .addPanel(
+      // Cortex (Thanos) doesn't track timing for 'iter', so we use ops/sec instead.
       g.panel('Op: Iter') +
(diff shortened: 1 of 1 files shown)


cortex-mixin/dashboards.libsonnet
```diff
@@ -50,8 +50,9 @@ local g = (import 'grafana-builder/grafana.libsonnet') + {
       g.latencyPanel(operationDuration, '{cluster=~"$cluster", operation="objectsize"}'),
     )
     .addPanel(
+      // Cortex (Thanos) doesn't track timing for 'iter', so we use ops/sec instead.
       g.panel('Op: Iter') +
-      g.latencyPanel(operationDuration, '{cluster=~"$cluster", operation="iter"}'),
+      g.queryPanel('sum(rate(%s%s[$__interval]))' % [opsTotal, '{cluster=~"$cluster", operation="iter"}'], 'ops/sec')
```
pracucci, line 55: I think the way you've written the query here is a bit complicated to read. If you look at other `queryPanel()` it's usually something like this:

```suggestion
      g.queryPanel('sum(rate(%s{cluster=~"$cluster", operation="iter"}[$__interval]))' % [opsTotal], 'ops/sec')
```

pstibrany: You're correct.

PR #41, opened 2020-04-15
Update config for internal r81 release (1.0.0)

gouthamve:

### DIFF
--- a/cortex/config.libsonnet
+++ b/cortex/config.libsonnet
@@ -37,7 +37,6 @@
     querier: {
       replicas: if $._config.queryFrontend.sharded_queries_enabled then 12 else 6,
       concurrency: if $._config.queryFrontend.sharded_queries_enabled then 16 else 8,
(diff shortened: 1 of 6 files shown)


cortex/config.libsonnet
```diff
@@ -165,18 +158,14 @@
...
         }
       else {}
     ),
 
     ringConfig: {
       'consul.hostname': 'consul.%s.svc.cluster.local:8500' % $._config.namespace,
-      'consul.consistent-reads': false,
       'ring.prefix': '',
-      'consul.watch-rate-limit': 1,
-      'consul.watch-burst-size': 5,
```
sandeepsukhani, line 179: The new default is now 1, is this change intentional?

gouthamve: Yes, it has been running in several prod envs without issues :) Its just the burst size, and if we are doing a qps more than 1, it'll be pretty quickly filled up anyways.

PR #43, opened 2020-04-19
Make the cortex-mixin work on non-k8s deployments.

tomwilkie:
But first, a bunch of refactoring to tidy the dashboards up:
- Put all the mixin config in one place.
- Make dashboardWithTagsAndLinks just an override on the dashboard constructor.
- Factor out the helper functions.
- Move the dashboards to separate files and namespace them off so they only have access to the config.
- Add a `AddRowIf` helper to massively rationalise the inclusion / exclusion of rows.
- Alter how we conditionally include dashboards themselves.
- Use set-style selector for bigtable/cassandra/dynamodb.
- Make the mixin by default not need any config, and generate dashboards for all possible permutations of tsdb, chunks, bigtable, dynamodb, cassandra, gcs, s3 etc.


cortex-mixin/config.libsonnet
```diff
@@ -0,0 +1,26 @@
...
+
+  _config+:: {
+    // Switch for overall storage engine.
+    // May contain 'chunks', 'tsdb' or both.
+    // Enables chunks- or tsdb- specific panels and dashboards.
+    storage_engine: ['chunks', 'tsdb'],
+
+    // For chunks backend, switch for chunk index type.
+    // May contain 'bigtable', 'dynamodb' or 'cassandra'.
+    chunk_index_backend: ['bigtable', 'dyamodb', 'cassandra'],
```
jtlisi, line 13:
```suggestion
    chunk_index_backend: ['bigtable', 'dynamodb', 'cassandra'],
```

cortex-mixin/config.libsonnet
```diff
@@ -0,0 +1,26 @@
...
+
+    // For chunks backend, switch for chunk store type.
+    // May contain 'bigtable', 'dynamodb', 'cassandra', 's3' or 'gcs'.
+    chunk_store_backend: ['bigtable', 'dyamodb', 'cassandra', 's3', 'gcs'],
```
jtlisi, line 17:
```suggestion
    chunk_store_backend: ['bigtable', 'dynamodb', 'cassandra', 's3', 'gcs'],
```

cortex-mixin/dashboards/dashboard-utils.libsonnet
```diff
@@ -0,0 +1,135 @@
...
+        if $._config.singleBinary
+        then self.addMultiTemplate('job', 'cortex_build_info', 'job')
+        else self
+             .addMultiTemplate('cluster', 'cortex_build_info', 'cluster')
+             .addMultiTemplate('namespace', 'cortex_build_info', 'namespace'),
+    },
+
+  // The ,ixin allow specialism of the job selector depending on if its a single binary
+  // deployment or a namespaced one.
+  jobMatcher(job)::
```
jtlisi, line 42: How are you going differentiate the metrics from each module using a Job? Is it going to be magic in the scrape config or are we going to allow cortex to run with component specific registries upstream?

tomwilkie: I'm not sure this is actually much of a problem - redundant metrics exported by jobs (ie ingester metrics from the distributor) don't influence the final results (ie qps or latency).  For almost all cases, we can target specific eg methods to get the qps/latency of a specific component.  In some cases (kv stores) the metrics aren't broken out like that yet[1], but for eg caches they are.  So I think we'll be okay.

Do you have a specific case in mind?

[1]https://github.com/cortexproject/cortex/issues/2484

tomwilkie:
Thanks for the feedback Jacob!

> I noticed some dashboards and the recording rules still have cluster hardcoded but I assume that will come later.

I think we can get away with still having the `cluster` label in recording rules, even when the underlying metric doesn't have a cluster label.  Its basically just ignored, and I prefer having consistent recording rules across deployments.  We might even have multiple clusters of single process Cortex, and re-introduce the cluster label...

cortex-mixin/dashboard-utils.libsonnet
```diff
@@ -0,0 +1,102 @@
+(import 'grafana-builder/grafana.libsonnet') {
+
+  // Override the dashboard constructor to add:
+  // - default tags,
+  // - some links that propagate the selectred cluster.
+  dashboard(title)::
```
pstibrany, line 6: Only dashboards using common set of datasource/cluster/namespace variables were using `dashboardWithTagsAndLinks`, but after this rename, all dashboards do. Links to such dashboards are broken and don't work correctly when navigating through them.

tomwilkie: @pstibrany good catch - I've made the links get added automatically if `addClusterSelectorTemplate` is called.

pstibrany: I think that should include tags. Links point to other dashboards with 'cortex' tag, so this tag(s) should only be added if there should links pointing to it.

PR #45, opened 2020-04-21
Update Ruler Storage Jsonnet

jtlisi:
This updates some of the flags used to set up the backend for running the ruler. It also adds support for the configdb, gcs, and S3.


jtlisi:
[Update cortex-jsonnet to use correct flags for the ruler](https://trello.com/c/EwYFjvJi/357-update-cortex-jsonnet-to-use-correct-flags-for-the-ruler)

PR #46, opened 2020-04-27
Add dashboard for WAL metrics

codesome:
/cc @tomwilkie @gouthamve

### DIFF
--- a/cortex-mixin/dashboards/chunks.libsonnet
+++ b/cortex-mixin/dashboards/chunks.libsonnet
@@ -50,4 +50,51 @@ local utils = import 'mixin-utils/utils.libsonnet';
         $.qpsPanel('cortex_ingester_chunk_age_seconds_count{%s}' % $.jobMatcher('ingester')),
       ),
     ),
+
+  'cortex-wal.json':
+    $.dashboard('Cortex / WAL')
+    .addClusterSelectorTemplates()
+    .addRow(
+      $.row('')
+      .addPanel(
+        $.panel('Bytes Logged (WAL+Checkpoint) / ingester / second') +
+        $.queryPanel('avg(rate(cortex_ingester_wal_logged_bytes_total{%(m)s}[$__interval])) + avg(rate(cortex_ingester_checkpoint_logged_bytes_total{%(m)s}[$__interval]))' % { m: $.jobMatcher('ingester') }, 'bytes') +
+        { yaxes: $.yaxes('bytes') },
+      )
+    )
+    .addRow(
+      $.row('WAL')
+      .addPanel(
+        $.panel('Records logged / ingester / second') +
+        $.queryPanel('avg(rate(cortex_ingester_wal_records_logged_total{%s}[$__interval]))' % $.jobMatcher('ingester'), 'records'),
+      )
+      .addPanel(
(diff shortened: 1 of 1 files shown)


cortex-mixin/dashboards.libsonnet
```diff
@@ -11,7 +11,7 @@
      else {}) +
 
     (if std.setMember('chunks', $._config.storage_engine)
-     then import 'dashboards/chunks.libsonnet'
+     then (import 'dashboards/chunks.libsonnet') + (import 'dashboards/wal.libsonnet')
```
gouthamve, line 14: Can we put this in `chunks.libsonnet` itself as the WAL is part of chunks?

codesome: Done

cortex-mixin/dashboards/chunks.libsonnet
```diff
@@ -50,4 +50,43 @@ local utils = import 'mixin-utils/utils.libsonnet';
...
+      .addPanel(
+        $.panel('Bytes per sample') +
+        $.queryPanel('avg(rate(cortex_ingester_wal_logged_bytes_total{cluster=~"$cluster", job=~"($namespace)/ingester"}[$__interval]) / rate(cortex_ingester_ingested_samples_total{cluster=~"$cluster", job=~"($namespace)/ingester"}[$__interval]))', 'bytes'),
+      )
+    )
+    .addRow(
+      $.row('Checkpoint')
+      .addPanel(
+        $.panel('Checkpoint creation/deletion / sec') +
+        $.queryPanel('rate(cortex_ingester_checkpoint_creations_total{cluster=~"$cluster", job=~"($namespace)/ingester"}[$__interval])', '{{instance}}-creation') +
```
gouthamve, line 83: I am not sure a per instance graph would be useful here

cortex-mixin/dashboards/chunks.libsonnet
```diff
@@ -50,4 +50,43 @@ local utils = import 'mixin-utils/utils.libsonnet';
...
+        $.queryPanel('rate(cortex_ingester_checkpoint_deletions_total{cluster=~"$cluster", job=~"($namespace)/ingester"}[$__interval])', '{{instance}}-deletion'),
+      )
+      .addPanel(
+        $.panel('Checkpoint creation/deletion failed / sec') +
+        $.queryPanel('rate(cortex_ingester_checkpoint_creations_failed_total{cluster=~"$cluster", job=~"($namespace)/ingester"}[$__interval])', '{{instance}}-creation') +
+        $.queryPanel('rate(cortex_ingester_checkpoint_deletions_failed_total{cluster=~"$cluster", job=~"($namespace)/ingester"}[$__interval])', '{{instance}}-deletion'),
+      )
+    ),
```
gouthamve, line 91: Can we add a panel that exports the min(free_space) on the mounted disk, etc.

cortex-mixin/dashboards/chunks.libsonnet
codesome: It is handy to click on a single instance and see when the checkpointing happened for it. If we have a sum of them, it doesn't seem to be that useful, or can you think of a use-case?

PR #47, opened 2020-04-28
Immediately join ingesters in the ring when using the blocks storage

pracucci:
When using the blocks storage we do disable blocks transfer on ingesters shutdown (because of the WAL on persistent disks) so I don't see a good reason to wait 30s before a new ingester joins the ring (is there?).

### DIFF
--- a/cortex/tsdb.libsonnet
+++ b/cortex/tsdb.libsonnet
@@ -82,6 +82,7 @@
   ingester_args+:: {
     // Disable TSDB blocks transfer because of persistent volumes
     'ingester.max-transfer-retries': 0,
+    'ingester.join-after': '0s',
 
     // Persist ring tokens so that when the ingester will be restarted
     // it will pick the same tokens


pracucci:
> Just 1 question: is disabling block transfer hardcoded or configurable?

It's configurable. It's this line:
```
'ingester.max-transfer-retries': 0,
```

_I agree it's a bit obscure, but that's how it works right now._

codesome:
If the aim is to always load from the disk, then the transfer can be disabled by default, along with join-after to 0 (similar to WAL). But for now, this PR LGTM.

pracucci:
> If the aim is to always load from the disk, then the transfer can be disabled by default, along with join-after to 0 (similar to WAL).

We do supports blocks transfer too in Cortex, reason why we don't have any enforcement in Cortex. It's the user/operator choice whether to use the blocks/WAL transfer or run ingesters on persistent disks.

PR #48, opened 2020-04-30
Make alert name consistent with others

codesome:


pstibrany:
Ganesh, can you please take a look and fix conflicts?

PR #52, opened 2020-05-06
Fixes #49

### LINKED ISSUE #49: Query ingesters within 12h too long?
I am not sure why it would make sense to set ` querier.query-ingesters-within` to 12h as shown here: https://github.com/grafana/cortex-jsonnet/blob/master/cortex/config.libsonnet#L149 .

As the comment says the ingesters store chunks 6h and the configured retain period is 15m (see https://github.com/grafana/cortex-jsonnet/blob/master/cortex/ingester.libsonnet#L21-L22). Doesn't that mean chunks can only live up to 6h15min in ingesters or is there something else to consider as well?

One might want to change this value or add a better comment why 12h is reasonable I think.

weeco:
Signed-off-by: Martin Schneppenheim <martin.schneppenheim@rewe-digital.com>

Thanks for the explanation @gouthamve and @tomwilkie , I thought it's worth to put that as comment right next to the code/config.


jtlisi:
@weeco would you mine rebasing your PR on top of master to fix the merge conflicts?

PR #57, opened 2020-05-07
Improved blocks storage support in the Reads and Queries dashboards

pracucci:
In this PR:
- Clarified which panels related only to "Chunks storage" and which to "Blocks storage" (our use case is that we have to enable both storage engines per Grafana installation)
- "Cortex / Reads" focus on the read-path
- "Cortex / Queries" focus on internal details about query execution
- Fixed namespace selector for object store stats


pracucci:
> I think querier part, where it talks about chunk corruptions is also chunk specific, but we can fix that when adding block chunk cache panels.

Yes, but in that row the "Stages" panel applies to blocks storage too, so I'm not 100% sure how we can address it easily. Let's discuss it and I will submit another PR for that.

PR #62, opened 2020-05-08
Added blocks storage alerts

pracucci:
Existing Cortex alerts already cover the write and read path of the blocks storage too, but there are few critical conditions not covered.

In this PR:
- Compactor has not successfully shipped blocks
- Compactor has not successfully completed a run
- Ingester is not successfully shipping blocks
- Querier has not successfully completed a bucket scan
- Store-gateway has not successfully completed a bucket sync

### DIFF
--- a/cortex-mixin/alerts.libsonnet
+++ b/cortex-mixin/alerts.libsonnet
@@ -2,5 +2,11 @@
   prometheusAlerts+::
     (import 'alerts/alerts.libsonnet') +
 
+    (if std.setMember('tsdb', $._config.storage_engine)
+     then
+       (import 'alerts/blocks.libsonnet') +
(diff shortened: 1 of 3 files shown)


pracucci:
> For compactor run, would it make sense to check for last compactor run time instead?

I would prefer to add both. The reason is that compactor should run successfully (and checking last timestamp is fine), but it's also expected to upload some blocks. If compactor runs successfully but no block is uploaded (ie. it could run but compact no blocks), then something weird is happening in the cluster.

pracucci:
@pstibrany I've added the missing alerts, given the Cortex PR https://github.com/cortexproject/cortex/pull/2573 has been merged now. Could you review it again, please?

PR #64, opened 2020-05-13
Added 'Writes Resources' and 'Reads Resources' dashboards

pracucci:
While investigating performance issues (ie. high latency), I frequently find myself starting from basic resources metrics like CPU and memory. In this PR I'm proposing to add two new dashboards:

- "Cortex / Write Resources" covering CPU / memory in the write path
- "Cortex / Read Resources" covering CPU / memory in the read path

### Previews

Screen Shot 2020-05-13 at 12 21 10

Screen Shot 2020-05-13 at 12 21 13

### DIFF
--- a/cortex-mixin/config.libsonnet
+++ b/cortex-mixin/config.libsonnet
@@ -25,5 +25,8 @@
 
     cortex_p99_latency_threshold_seconds: 2.5,
     alert_namespace_matcher: '',
+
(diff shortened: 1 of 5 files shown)


pracucci:
> This looks awesome, but it's specific to environments with cAdvisor. I'm wondering if it's worth having behind a flag. Other than that, I'm really looking forward to have this available!

💯 agree. I've added a config option to enable it (disabled by default).

PR #70, opened 2020-05-18
Update instructions in README

pstibrany:
Building dashboards needs -m parameter.


pstibrany:
Perhaps we should just ask people to use `make`?

PR #71, opened 2020-05-21
Enable chunks caching when using blocks

pstibrany:


cortex/config.libsonnet
```diff
@@ -64,10 +64,12 @@
     // TSDB storage engine doesn't require the table manager.
     table_manager_enabled: $._config.storage_engine != 'tsdb',
 
-    // TSDB storage engine doesn't require memcached for chunks or chunk indexes.
+    // TSDB storage engine doesn't support memcached for chunk indexes.
     memcached_index_queries_enabled: $._config.storage_engine != 'tsdb',
     memcached_index_writes_enabled: $._config.storage_engine != 'tsdb',
-    memcached_chunks_enabled: $._config.storage_engine != 'tsdb',
+
+    // In latest Cortex chunks cache is now supported by both TSDB storage engine and chunks engine.
```
pracucci, line 71:
```suggestion
    // Chunks cache is supported by both TSDB storage engine and chunks engine.
```

cortex/tsdb.libsonnet
```diff
@@ -23,6 +23,13 @@
     cortex_compactor_data_disk_class: 'standard',
   },
 
+  blocks_chunks_caching_config:: {
+    'experimental.tsdb.bucket-store.chunks-cache.backend': 'memcached',
+    'experimental.tsdb.bucket-store.chunks-cache.memcached.addresses': 'dnssrvnoa+memcached.%(namespace)s.svc.cluster.local:11211' % $._config,
+    'experimental.tsdb.bucket-store.chunks-cache.memcached.timeout': '200ms',
+    'experimental.tsdb.bucket-store.chunks-cache.memcached.max-item-size': 5 * 1024 * 1024,
```
pracucci, line 30: This should be guaranteed to be consistent with the memcached config. The memcached default is 1mb, and we don't override it for the chunks cache, so we should set it to `1 * 1024 * 1024` here as well, but we have a shared variable between the two we can guarantee they're aligned.

pstibrany: Done. Please take a look.

pstibrany:
Also added index_queries cache setup.

cortex/config.libsonnet
```diff
@@ -64,10 +64,16 @@
     // TSDB storage engine doesn't require the table manager.
     table_manager_enabled: $._config.storage_engine != 'tsdb',
 
-    // TSDB storage engine doesn't require memcached for chunks or chunk indexes.
-    memcached_index_queries_enabled: $._config.storage_engine != 'tsdb',
+    // TSDB storage engine doesn't support index-writes (for writes deduplication) cache.
     memcached_index_writes_enabled: $._config.storage_engine != 'tsdb',
-    memcached_chunks_enabled: $._config.storage_engine != 'tsdb',
+    memcached_index_writes_max_item_size_mb: 5,
```
pracucci, line 69: The default here should be 1.

cortex/memcached.libsonnet
```diff
@@ -33,21 +33,23 @@ memcached {
...
-      max_item_size: '5m',
+      max_item_size: '%dm' % [$._config.memcached_index_queries_max_item_size_mb],
     }
   else {},
 
   // Dedicated memcached instance used to dedupe writes to the index.
   memcached_index_writes: if $._config.memcached_index_writes_enabled then
     $.memcached {
       name: 'memcached-index-writes',
+      max_item_size: '%dm' % [$._config.memcached_index_queries_max_item_size_mb],
```
pracucci, line 44:
```suggestion
      max_item_size: '%dm' % [$._config.memcached_index_writes_max_item_size_mb],
```

cortex/tsdb.libsonnet
```diff
@@ -23,6 +23,19 @@
     cortex_compactor_data_disk_class: 'standard',
   },
 
+  blocks_chunks_caching_config:: {
+    'experimental.tsdb.bucket-store.index-cache.backend': 'memcached',
+    'experimental.tsdb.bucket-store.index-cache.memcached.addresses': 'dnssrvnoa+memcached-index-queries.%(namespace)s.svc.cluster.local:11211' % $._config,
+    'experimental.tsdb.bucket-store.index-cache.memcached.timeout': '200ms',
+    'experimental.tsdb.bucket-store.index-cache.postings-compression-enabled': 'true',
+    'experimental.tsdb.bucket-store.index-cache.memcached.max-item-size': $._config._memcached_index_queries_max_item_size_mb * 1024 * 1024,
```
pracucci, line 31: I would also set better defaults for queue/batch:
```
'experimental.tsdb.bucket-store.index-cache.memcached.max-async-buffer-size': '25000',
    'experimental.tsdb.bucket-store.index-cache.memcached.max-async-concurrency': '50',
    'experimental.tsdb.bucket-store.index-cache.memcached.max-get-multi-batch-size': '100',
```

pstibrany: I left them out as we use this extra config in some envs, but not all. I will add it.

PR #72, opened 2020-05-21
Add jsonnet for ingester StatefulSet with WAL

codesome:


codesome:
> I'm just wondering if we should put it behind a flag, so that people using this can easily turn it off if they are not ready to migrate yet.

I was thinking this repo is like a reference and our suggested way of running Cortex. If it's a new user, then they should always run with WAL ideally. If someone wants to migrate, they could get a statefulset up using this as reference and follow the [guide for migration](https://cortexmetrics.io/docs/production/ingesters-with-wal/#migrating-from-stateless-deployments), and get rid of the deployment. WDYT?

gouthamve:
We should support both deployment and statefulset for now

forestsword:
What's the answer to this look like now that it's August? I'd like us to start using the libraries here and we're through the migration from deployment to a statefulset.

codesome:
@forestsword I will be cleaning this up tomorrow. We will be supporting both the deployment without WAL and statefulset with WAL.

cortex/ingester.libsonnet
```diff
@@ -40,9 +38,28 @@
...
+    'ingester.tokens-file-path': $._config.wal_dir + '/tokens',
+  },
+
+  _config+:: {
+    wal_dir: '/wal_data',
+    ingester+: {
+      statefulset_replicas: 3,
+      statefulset_disk: '150Gi',
+    },
+  },
```
pstibrany, line 57: Should this go directly to config.libsonnet?

pstibrany (approved):
Looks good to me.

cortex/ingester.libsonnet
codesome: Yes :)

codesome: I also moved `wal_dir` inside `_config.ingester` where it made most sense. Will merge on green.

PR #76, opened 2020-05-25
Make job names configurable

jtlisi:
This PR allows the operator to configure the job names for the generated dashboards. This allows users to configure the mixin for non-default deployments of cortex and configure the mixin for use with the single binary deployment of cortex. A single binary deployment of cortex will not have unique job s for all of its modules. Instead, `$namespace/cortex` will be the job name in most deployments. Exposing the job constant used in the cortex dashboards will allow the user to configure and generate dashboards that work best for them.

For example:
Assuming a default micro-service deployment of cortex with the standard job names and a single binary deployment with the job name `cortex`. The following config could be set to allow for dashboards that accommodate both deployments.

```jsonnet
{
  _config: {
    job_names: {
      ingester: '(ingester|cortex$)',
      distributor: ('distributor|cortex$)',
      querier: '(querier|cortex$)',
      query_frontend: '(query-frontend|cortex$)',
      table_manager: '(table-manager|cortex$)',
      store_gateway: '(store-gateway|cortex$)',
    },
  }
}
```

- Later this can ideally be simplified into a smaller selection of configs

### DIFF
--- a/cortex-mixin/config.libsonnet
+++ b/cortex-mixin/config.libsonnet
@@ -23,6 +23,16 @@
     // modify the job selectors in the dashboard queries.
     singleBinary: false,
 
+    job_names: {
+      ingester: 'ingester',
+      distributor: 'distributor',
+      querier: 'querier',
+      query_frontend: 'query-frontend',
+      table_manager: 'table-manager',
+      store_gateway: 'store-gateway',
+      gateway: 'cortex-gw',
+    },
+
     cortex_p99_latency_threshold_seconds: 2.5,
     alert_namespace_matcher: '',
 
--- a/cortex-mixin/dashboards/chunks.libsonnet
+++ b/cortex-mixin/dashboards/chunks.libsonnet
@@ -8,46 +8,46 @@ local utils = import 'mixin-utils/utils.libsonnet';
       $.row('Active Series / Chunks')
       .addPanel(
         $.panel('Series') +
-        $.queryPanel('sum(cortex_ingester_memory_series{%s})' % $.jobMatcher('ingester'), 'series'),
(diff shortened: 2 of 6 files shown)


tomwilkie:
Looks like your trying to make this work for the single binary - does my previous effort not work here?

tomwilkie:
Looks like you’re trying to make this work for the single binary - does my previous PR not do the trick?

jtlisi:
@cstyan I deployed it to the development cluster and it worked pretty well there. I want to deploy it again and double check.

@tomwilkie We are still running into two issues. First, is that the job names in the dashboard are still set to `ingester`/`distributor`/etc. When we run a single binary internally the job name for the scrape will be `cortex`. Second, internally we run both microservices and single binary deployments. Which means our dashboards need to somehow accommodate both. By exposing the job names as variables we can inject a regex that works for both microservices or singleBinary. However, the config would be much simple if we just deployed one or the other.

tomwilkie:
Can we generate & install two version of the mixin then?

The challenge with this approach is it won't work for people not on k8s.  And I'd rather avoid have 3 versions of the mixin if possible.

jtlisi:
@tomwilkie For people not on k8s they can still set the `singleBinary` flag to true and that will remove label selectors for `cluster` and `namespace` and just leave the selectors for the job names. So if someone was running on bare metal they could just set that flag and update all of the job names to be `cortex` if they are running on a single binary or leave them as is if they are running microservices on baremetal. The `singleBinary` config option acts more as a `enableK8sLabels` config option.

jtlisi:
> LGTM, have you tried this in QA or anywhere to ensure those dashboards still work?
> 
> and for your example `component|cortex$`, the `$` is jsonnet specific syntax?

The `$` is just to ensure the regex terminates after `cortex` otherwise the `cortex-gw` would always match as well.

PR #77, opened 2020-05-26
Split recording rules into separate groups

jtlisi:
The size of the rule group for recording rules in this mixin is getting a bit large. It's currently taking `>15s` to run on a standard prometheus: 

image

This PR breaks the recording rules for cortex into smaller rule groups which provide a bit more wiggle room for anyone with low eval intervals on their recording rules to avoid missed iterations.

### DIFF
--- a/cortex-mixin/recording_rules.libsonnet
+++ b/cortex-mixin/recording_rules.libsonnet
@@ -15,7 +15,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
         rules:
           utils.histogramRules('cortex_memcache_request_duration_seconds', ['cluster', 'job', 'method']) +
           utils.histogramRules('cortex_cache_request_duration_seconds', ['cluster', 'job']) +
-          utils.histogramRules('cortex_cache_request_duration_seconds', ['cluster', 'job', 'method'])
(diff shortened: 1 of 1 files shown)


cortex-mixin/recording_rules.libsonnet
```diff
@@ -4,22 +4,25 @@ local utils = import 'mixin-utils/utils.libsonnet';
...
+      {
+        name: 'cortex_cache',
+        rules:
           utils.histogramRules('cortex_memcache_request_duration_seconds', ['cluster', 'job', 'method']) +
           utils.histogramRules('cortex_cache_request_duration_seconds', ['cluster', 'job']) +
-          utils.histogramRules('cortex_cache_request_duration_seconds', ['cluster', 'job', 'method']) +
+          utils.histogramRules('cortex_cache_request_duration_seconds', ['cluster', 'job', 'method']),
+      },
+      {
+        name: 'cortex_chunk_store',
```
gotjosh, line 21:
```suggestion
        name: 'cortex_storage',
```

gotjosh: Given we have things like `kv` and `gcs` which are used by both.

jtlisi: makes sense

PR #78, opened 2020-05-27
Configure metadata cache when using blocks storage.

pstibrany:
Cache is used by both querier (even if store-gateway is disabled) and store-gateway.

### DIFF
--- a/cortex/tsdb.libsonnet
+++ b/cortex/tsdb.libsonnet
@@ -42,6 +42,16 @@
     'experimental.tsdb.bucket-store.chunks-cache.memcached.max-get-multi-batch-size': '100',
   },
 
+  blocks_metadata_caching_config:: {
+    'experimental.tsdb.bucket-store.metadata-cache.backend': 'memcached',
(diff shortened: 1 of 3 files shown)


cortex/tsdb.libsonnet
```diff
@@ -189,7 +199,7 @@
       // Persist ring tokens so that when the store-gateway will be restarted
       // it will pick the same tokens
       'experimental.store-gateway.tokens-file-path': '/data/tokens',
-    } + (if $._config.store_gateway_enabled then $.blocks_chunks_caching_config else {}),
+    } + $.blocks_chunks_caching_config + $.blocks_metadata_caching_config,
```
pracucci, line 202: Chunks caching config should be specified only if store gateway is **disabled** or am I wrong?

pstibrany: This is `store_gateway_args`, if store gateway is disabled, this doesn't get used.

pracucci: Damn, I got confused during a quick look at the PR. Sorry!

pstibrany: No worries, thanks for your review!

PR #82, opened 2020-05-29
Add alert for TLS cert expiration

annanay25:
Signed-off-by: Annanay <annanayagarwal@gmail.com>

### DIFF
--- a/cortex-mixin/alerts/alerts.libsonnet
+++ b/cortex-mixin/alerts/alerts.libsonnet
@@ -461,5 +461,26 @@
         },
       ],
     },
+    {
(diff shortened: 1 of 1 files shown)


pstibrany:
Which component generates `ssl_cert_expiry_timestamp_seconds` metric? I cannot find it in Cortex (Is that correct?), which would imply that alert should not be in public jsonnet.

jtlisi:
I'm going to close this for now since there is no plan to export the `ssl_cert_expiry_timestamp_seconds` metric in cortex. See https://github.com/cortexproject/cortex/pull/2592

PR #88, opened 2020-06-04
Improved Cortex blocks compactor alerts and dashboard

pracucci:
In this PR I'm proposing few changes related to the blocks compactor:
1. Dashboard: remove panel displaying "per-tenant compactions" (based on `cortex_compactor_group_compactions_*` metrics) because grouping logic is not necessarily "1 group per tenant"
2. Alert: remove alert on `cortex_compactor_block_cleanup_last_successful_run_timestamp_seconds` because, as of today, this metric can get never updated if the compactor is always busy doing compaction. This is due to how Thanos `BucketCompactor` works (the `Compact()` function doesn't return as far as there's some work to do). I will work to propose an improvement to Thanos to have a metric there from which we can alert on.
3. Alert: add an alert if the blocks cleanup didn't successfully run (this is a recent introduction in Cortex, because it was coupled to the blocks compaction loop).

## Preview of the compactor dashboard

Screen Shot 2020-06-04 at 17 10 34

### DIFF
--- a/cortex-mixin/alerts/compactor.libsonnet
+++ b/cortex-mixin/alerts/compactor.libsonnet
@@ -4,33 +4,33 @@
       name: 'cortex_compactor_alerts',
       rules: [
         {
-          // Alert if the compactor has not successfully completed a run in the last 24h.
-          alert: 'CortexCompactorHasNotSuccessfullyRun',
+          // Alert if the compactor has not successfully cleaned up blocks in the last 24h.
+          alert: 'CortexCompactorHasNotSuccessfullyCleanedUpBlocks',
           'for': '15m',
           expr: |||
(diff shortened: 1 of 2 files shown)


pracucci:
> Alert: remove alert on cortex_compactor_block_cleanup_last_successful_run_timestamp_seconds because, as of today, this metric can get never updated if the compactor is always busy doing compaction. This is due to how Thanos BucketCompactor works (the Compact() function doesn't return as far as there's some work to do). I will work to propose an improvement to Thanos to have a metric there from which we can alert on.

I forgot we also have an alert on the uploaded blocks, so we're partially covered anyway (at least for the case "compactor is not running at all"). The alert on uploaded blocks is also the way Thanos alerting works on their mixin.

PR #91, opened 2020-06-11
Refactor Alerts to work with single-binary job

jtlisi:
Since Cortex has adopted embedding metrics within specific components, the `job` label is no longer needed on a number of alerts. Removing a hardcoded `job` label makes it easier to reuse these rules in single-binary and non-k8s deployments.

### DIFF
--- a/cortex-mixin/alerts/alerts.libsonnet
+++ b/cortex-mixin/alerts/alerts.libsonnet
@@ -7,7 +7,7 @@
           alert: 'CortexIngesterUnhealthy',
           'for': '15m',
           expr: |||
-            min(cortex_ring_members{state="Unhealthy", job=~"[a-z]+/distributor" %s}) by (namespace, job) > 0
+            min(cortex_ring_members{state="Unhealthy" %s}) by (namespace, job) > 0
           ||| % $.namespace_matcher(','),
           labels: {
             severity: 'critical',
@@ -120,7 +120,7 @@
         {
           alert: 'CortexQuerierCapacityFull',
           expr: |||
-            prometheus_engine_queries_concurrent_max{job=~".+/querier"} - prometheus_engine_queries{job=~".+/querier"} == 0
+            prometheus_engine_queries_concurrent_max - prometheus_engine_queries == 0
           |||,
           'for': '5m',  // We don't want to block for longer.
           labels: {
@@ -135,7 +135,7 @@
         {
           alert: 'CortexFrontendQueriesStuck',
           expr: |||
-            sum by (namespace) (cortex_query_frontend_queue_length{job=~".+/query-frontend" %s}) > 1
+            sum by (namespace) (cortex_query_frontend_queue_length{%s}) > 1
           ||| % $.namespace_matcher(','),
           'for': '5m',  // We don't want to block for longer.
           labels: {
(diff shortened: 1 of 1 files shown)


cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -120,7 +120,7 @@
         {
           alert: 'CortexQuerierCapacityFull',
           expr: |||
-            prometheus_engine_queries_concurrent_max{job=~".+/querier"} - prometheus_engine_queries{job=~".+/querier"} == 0
+            prometheus_engine_queries_concurrent_max - prometheus_engine_queries == 0
```
jtlisi, line 123: This change allows a ruler service to also trigger this alert since it depends on the querier module. That is ok since this alert is important for operating the ruler as well.

cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -7,7 +7,7 @@
           alert: 'CortexIngesterUnhealthy',
           'for': '15m',
           expr: |||
-            min(cortex_ring_members{state="Unhealthy", job=~"[a-z]+/distributor" %s}) by (namespace, job) > 0
+            min(cortex_ring_members{state="Unhealthy", name="ingester" %s}) by (namespace, job) > 0
```
jtlisi, line 10: We can use the ring name label instead of the distributor for this alert.

cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -135,7 +135,7 @@
         {
           alert: 'CortexFrontendQueriesStuck',
           expr: |||
-            sum by (namespace) (cortex_query_frontend_queue_length{job=~".+/query-frontend" %s}) > 1
+            sum by (namespace) (cortex_query_frontend_queue_length{%s}) > 1
```
jtlisi, line 138: This metric is now only reported by the `query-frontend` module

cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -449,7 +449,7 @@
           expr: |||
             memberlist_client_cluster_members_count{%s}
               != on (cluster,namespace) group_left
-            sum(up{job=~".+/(distributor|ingester|querier)"}) by (cluster,namespace)
+            sum(up{job=~".+/(distributor|ingester|querier|cortex|ruler)"}) by (cluster,namespace)
```
jtlisi, line 452: All jobs that use a ring need to be included

cortex-mixin/alerts/alerts.libsonnet
pracucci: My main concern is that without any label selector it also picks up Prometheus, which is misleading (and also wrong if different teams are on-call for Prometheus and Cortex).

jtlisi: In that case, one option would be using a pattern of multiple job names for any component that includes a Prometheus engine. Similar to to the memberlist alert below.

cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -120,7 +120,7 @@
         {
           alert: 'CortexQuerierCapacityFull',
           expr: |||
-            prometheus_engine_queries_concurrent_max{job=~".+/querier"} - prometheus_engine_queries{job=~".+/querier"} == 0
+            prometheus_engine_queries_concurrent_max{job=~".+/(cortex|ruler|querier)"}, - prometheus_engine_queries{job=~".+/(cortex|ruler|querier)"}, == 0
```
pracucci, line 123: Using the `job` selector LGTM, but the query is wrong now

jtlisi: Good catch. I updated the query to remove the commas

PR #99, opened 2020-06-18
Fix selector in querier service

metost:
The selector should be "querier" instead of "query-frontend".
Please correct me if I'm wrong.


pracucci (commented):
Thanks for your PR. We're aware of this and I agree we should fix it. @jtlisi I think at some point you planned to take a look at it too?

tleopold:

PR #100, opened 2020-06-19
Added blocks storage refetch metrics and alerts

pracucci:
I've just merged the Cortex PR https://github.com/cortexproject/cortex/pull/2695 that introduced refetching of series from a different store-gateway in case of missing blocks (consistency check failed).

_In this PR I'm introducing some new metrics and alerts related to that change._

### DIFF
--- a/cortex-mixin/alerts/blocks.libsonnet
+++ b/cortex-mixin/alerts/blocks.libsonnet
@@ -50,6 +50,30 @@
           },
         },
         {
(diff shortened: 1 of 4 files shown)


pracucci:
> LGTM, but I'm not sure about change for `cortex_cache_hits` -- why is `name=~"frontend.+"` removed there?

Thanks for spotting it! It was an unintended change.

PR #109, opened 2020-06-25
Add config option to disable HA tracker

amckinley:
I don't want to create an etcd cluster at all because of https://github.com/grafana/jsonnet-libs/issues/276

The HA tracker appears to be the only use for etcd, so add a flag to disable it, thus removing the etcd dependency entirely.

### DIFF
--- a/cortex/config.libsonnet
+++ b/cortex/config.libsonnet
@@ -195,6 +195,8 @@
       'ring.prefix': '',
     },
 
+    enable_ha_tracker: true,
+
     // Some distributor config is shared with the querier.
     distributorConfig: {
       'distributor.replication-factor': $._config.replication_factor,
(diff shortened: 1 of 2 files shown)


tomwilkie:
Hi Austin! I think all you need to do is specify the following in your config to disable this.  The extra flags will be a noop if this is set.

```
{
  distributor_args+:: {
    'distributor.ha-tracker.enable': false,
  },

  etcd: {},
}
```

In general the reason we're using jsonnet is to avoid the extra "plumbing through" of config that you see in other systems (like helm).

tomwilkie:
(ps I've set the CI to run on forks now, so lint should work for you in the future)

amckinley:
Thanks @tomwilkie, that makes sense. How about a version of this PR that creates a config option to use either consul or etcd as the KV store for the HA tracker?

gouthamve:
That would be great @amckinley!

PR #110, opened 2020-06-25
Version bump to v1.1.0

amckinley:
Maybe this should be `1.2.0-rc0`?


gouthamve:
Lets make it 1.2.0 once the stable release is out. For now 1.1 is better!

PR #111, opened 2020-06-25
Use v1 API for Deployment and StatefulSet resources

amckinley:
AFAICT, there's no reason for these to be stuck to the `apps/v1beta1` API version. Note that I kept the `v1beta1` version for the `PodDisruptionBudget` that gets created.


tomwilkie:
(Going to re-push to get CI to run)

PR #112, opened 2020-06-25
Actually include the ruler

amckinley:
Not sure why this wasn't included in the example... Adds a dependency on this library and sets the default config similarly to the existing defaults.

Also updates the name of a deprecated config option.

### DIFF
--- a/cortex/config.libsonnet
+++ b/cortex/config.libsonnet
@@ -181,7 +181,7 @@
           // Setting for index cache.
           'store.index-cache-validity': '14m',  // ingester.retain-period=15m, 1m less for safety.
           'store.index-cache-read.cache.enable-fifocache': true,
-          'store.index-cache-read.fifocache.size': 102400,
+          'store.index-cache-read.fifocache.max-size-items': 102400,
           'store.index-cache-read.memcached.hostname': 'memcached-index-queries.%(namespace)s.svc.cluster.local' % $._config,
           'store.index-cache-read.memcached.service': 'memcached-client',
           'store.index-cache-read.memcached.timeout': '500ms',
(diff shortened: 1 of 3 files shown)


cortex/config.libsonnet
```diff
@@ -181,7 +181,7 @@
           // Setting for index cache.
           'store.index-cache-validity': '14m',  // ingester.retain-period=15m, 1m less for safety.
           'store.index-cache-read.cache.enable-fifocache': true,
-          'store.index-cache-read.fifocache.size': 102400,
+          'store.index-cache-read.fifocache.max-size-items': 102400,
```
pracucci, line 184: Good catch. `-store.index-cache-read.fifocache.size` is deprecated in favour of `-store.index-cache-read.fifocache.max-size-items` 👍

cortex/cortex.libsonnet
```diff
@@ -11,6 +11,7 @@
 (import 'querier.libsonnet') +
 (import 'query-frontend.libsonnet') +
 (import 'table-manager.libsonnet') +
+(import 'ruler.libsonnet') +
```
pracucci, line 14: I think we're not ready for this yet, unless we introduce a flag to conditionally enable/disable it. @jtlisi ?

gouthamve: I think we should be now. @gotjosh can answer better. And if we're not, we should fix it.

gouthamve: After a brief discussion with Josh, we're happy to have our users run ruler by default!

gotjosh: I think it's fine, the ruler appears to be in a good state. I do agree with Marco,  given this is an "optional" service we should make it trivial for our users to disable it.

tomwilkie:
(rebase&push to trigger CI)

PR #118, opened 2020-06-30
Added publish not ready addresses

joe-elliott:
Due to these changes https://github.com/cortexproject/cortex/pull/2733 we need to make adjustments to the way we configure our query frontend service.

Currently the queriers use the headless `query-frontend` service to discover frontends.  After the above change frontends will no longer be included in this service until their readiness probe returns 200.  This change will cause k8s to include query frontends in the `query-frontend` service that are not yet ready.  This will allow queriers to find them successfully.

This change can be rolled out before or alongside the reference cortex change.

### DIFF
--- a/cortex/querier.libsonnet
+++ b/cortex/querier.libsonnet
@@ -19,7 +19,7 @@
 
       // Limit to N/2 worker threads per frontend, as we have two frontends.
       'querier.worker-parallelism': $._config.querier.concurrency / $._config.queryFrontend.replicas,
-      'querier.frontend-address': 'query-frontend.%(namespace)s.svc.cluster.local:9095' % $._config,
+      'querier.frontend-address': 'query-frontend-discovery.%(namespace)s.svc.cluster.local:9095' % $._config,
       'querier.frontend-client.grpc-max-send-msg-size': 100 << 20,
 
       'log.level': 'debug',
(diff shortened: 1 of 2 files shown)


tomwilkie:
@joe-elliott whilst you in this space, can you tidy up the whole `querier` service?  Having that point to the query-frontends, and not the quieriers, is confusing.  Ideally we'd have a load-balanced service called query-frontend that the cortex-gw forwards to, a service called (perhaps) query-frontend-discovery which the queriers use to discover the query-frontend, and no querier service.  WDYT?  Maybe an ops task for this...

joe-elliott:
Will do.  I was confused by that initially.  This will require a bit of a manual touch on rollout, but nothing too terrible.

joe-elliott:
Ok, with the current changes we'd have these services:
- query-frontend
  - clusterIP service meant to take the query path
- query-frontend-discovery
  - headless clusterIP service meant for discovery from the queriers
- querier
  - clusterIP service that points to all queriers.  it seems this could be useful if we were having an issue with the query frontends where we could point directly at queriers if needed.  Will remove if we prefer that.

tomwilkie:
+1 be careful rolling this out.  Might want to leave the querier service pointing at the QF for now, and tidy that up in a second phase.

PR #106, opened 2020-06-23
Correct alerts to work with Cortex single binary

tomwilkie:
Builds on https://github.com/grafana/cortex-jsonnet/pull/102; opening a new PR to the the CI working.

Some big changes here:
- Specify the labels which can be used to get aggregate per-cluster metrics.  Ie for our deployments, those labels would be `(cluster, namespace)` but for a single-binary setup it would be `(job)`.
- Don't bother with functions, we just need to know a list of labels to aggregate by that will identify a single Cortex cluster (ie "cluster, namespace" or "job").
- Remove alert_namespace_matcher, its brittle and hard to maintain.
- Put WAL alerts in a separate group.
- Remove `CortexFlushStuck`, its been superseded by **CortexOldChunkInMemory**

PR #119, opened 2020-06-30
Added compactor to memberlist client

joe-elliott:
This is causing alerts for other elements of our infrastructure that happen to publish this metric as well.  The proposed change is odd, but would technically fix the issue for us.

@pstibrany I'm not super familiar with these alerts is there a better fix here?  Perhaps a way to add a filter to the alert or maybe prefix the `memberlist_` metrics for my application?

### DIFF
--- a/cortex-mixin/alerts/alerts.libsonnet
+++ b/cortex-mixin/alerts/alerts.libsonnet
@@ -432,7 +432,7 @@
           expr: |||
             memberlist_client_cluster_members_count
               != on (%s) group_left
-            sum by (%s) (up{job=~".+/(distributor|ingester|querier|cortex|ruler)"})
+            sum by (%s) (up{job=~".+/(distributor|ingester|querier|cortex|ruler|compactor)"})
           ||| % [$._config.alert_aggregation_labels, $._config.alert_aggregation_labels],
           'for': '5m',
           labels: {


pstibrany:
Compactor has nothing with memberlist, so I would like to check this before merging.

cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -432,7 +432,7 @@
           expr: |||
             memberlist_client_cluster_members_count
               != on (%s) group_left
-            sum by (%s) (up{job=~".+/(distributor|ingester|querier|cortex|ruler)"})
+            sum by (%s) (up{job=~".+/(distributor|ingester|querier|cortex|ruler|compactor)"})
```
pracucci, line 435: The only matching job I can find is `tracing-ops/compactor`. Is this the case you wanna cover? Could you share more details about the issue you're fixing, please?

pstibrany:
OK, in theory compactor could use memberlist and we should therefore have extra alert for that, but this is really solving a problem introduced by https://github.com/grafana/cortex-jsonnet/pull/106. Tempo-ops is reusing memberlist component and exporting metric, but it's not a Cortex environment, so Cortex alerts shouldn't be applied to it :(

pstibrany:
That said, I don't have a better suggestion.

joe-elliott:
Closing after discussing internally.  This change would actually break Cortex alerts b/c the Cortex compactor does not participate in memberlist.

PR #120, opened 2020-07-01
Update restarts alert, fixes #117.

### LINKED ISSUE #117: CortexIngesterRestarts
Right now `CortexIngesterRestarts` alert is noisy. It looks like this:

```
rate(process_start_time_seconds{job=~".+(cortex|ingester)"}[30m]) > 0
```

Rate over 30 minutes produces values like ~300 even for a single restart:

Screenshot 2020-06-30 at 14 04 28

We could use `changes` instead. Should we increase number of restarts (using `changes`) to be higher to 1 before alerting?

tomwilkie:

### DIFF
--- a/cortex-mixin/alerts/alerts.libsonnet
+++ b/cortex-mixin/alerts/alerts.libsonnet
@@ -160,13 +160,13 @@
         {
           alert: 'CortexIngesterRestarts',
           expr: |||
-            rate(process_start_time_seconds{job=~".+(cortex|ingester)"}[30m]) > 0
+            increase(process_start_time_seconds{job=~".+(cortex|ingester)"}[30m]) > 1
           |||,
           labels: {
             severity: 'critical',
           },
           annotations: {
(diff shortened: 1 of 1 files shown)


cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -160,13 +160,13 @@
         {
           alert: 'CortexIngesterRestarts',
           expr: |||
-            rate(process_start_time_seconds{job=~".+(cortex|ingester)"}[30m]) > 0
+            increase(process_start_time_seconds{job=~".+(cortex|ingester)"}[30m]) > 1
```
beorn7, line 163: s/increase/changes/

It will fire too often with `increase` because the start time certainly increases by more than one after any restart. `changes` just counts how often the value has changed, so it gives you the actual number of restarts.

beorn7: #123

PR #123, opened 2020-07-01
Fix CortexIngesterRestarts alert

beorn7:


gouthamve:
Okay, why will this not trigger on a rollout?

beorn7:
It will trigger if you rollout more than once in 30m.

beorn7:
You could do something like "more than 5 restarts in 1h" to be even less noisy, but let's see how we fare with the current threshold.

pracucci:
_Thanks for fixing it! I should have noticed it during the review 🤦 _

PR #131, opened 2020-07-06
use single binary for job labels

jtlisi:
- Ensure the job label for each dashboard uses the default job label for microservice mode and single binary mode.


cstyan:
Does this mean dashboards/alerts using job selectors will return data for `component|cortex`, or do we only need to worry about that for environments with cortex deployed as both single binary and microservices?

jtlisi:
@cstyan this change allows the same set of dashboards and alerts to run for both single-binary and microservice setups simultaneously. We have been running this setup in our dev cluster for a month now and it works as expected

PR #137, opened 2020-07-13
Improved Cortex blocks storage playbooks

pracucci:
During the weekend we had an outage in the Cortex blocks storage. I've updated the playbooks based on the lessons learned.

### DIFF
--- a/cortex-mixin/docs/playbooks.md
+++ b/cortex-mixin/docs/playbooks.md
@@ -148,7 +148,7 @@ level=error ts=2020-07-12T17:35:05.516823471Z caller=compactor.go:339 component=
 ```
 
 When this happen you should:
-1. Rename the block prefixing it with `corrupted-` so that it will be skipped by the compactor and queriers
+1. Rename the block prefixing it with `corrupted-` so that it will be skipped by the compactor and queriers. Keep in mind that doing so the block will become invisible to the queriers too, so its series/samples will not be queried. It's safe to do it on a single block with compaction level 1 (because of the samples replication), but not on multiple overlapping level 1 blocks or any block with a compaction level > 1.
 2. Ensure the compactor has recovered
 3. Investigate offline the root cause (eg. download the corrupted block and debug it locally)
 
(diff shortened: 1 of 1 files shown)


cortex-mixin/docs/playbooks.md
```diff
@@ -130,6 +139,45 @@ How to investigate:
...
+### Compactor is failing because of `not healthy index found`
+
+The compactor may fail to compact blocks due a corrupted block index found in one of the source blocks:
+
+```
+level=error ts=2020-07-12T17:35:05.516823471Z caller=compactor.go:339 component=compactor msg="failed to compact user blocks" user=REDACTED err="compaction: group 0@6672437747845546250: block with not healthy index found /data/compact/0@6672437747845546250/REDACTED; Compaction level 1; Labels: map[__org_id__:REDACTED]: 1/1183085 series have an average of 1.000 out-of-order chunks: 0.000 of these are exact duplicates (in terms of data and time range)"
+```
+
+When this happen you should:
+1. Rename the block prefixing it with `corrupted-` so that it will be skipped by the compactor and queriers
```
pstibrany, line 151: This is only safe for 2h blocks from ingesters, because samples are replicated in other blocks. If bigger block is corrupted, renaming it will make it "invisible" for queries. Not sure what the procedure should be in such case. Perhaps we need a flag (marker file?) for compactor to ignore such block.

pracucci: I propose to:
1. Improve this playbook to mention what you're talking about (but keep in mind that if a block is corrupted, there's no guarantee querying that block will be a good thing at all)
2. Further discuss the marker here: https://github.com/cortexproject/cortex/issues/2866

pstibrany: SGTM

pracucci: > Improve this playbook to mention what you're talking about (but keep in mind that if a block is corrupted, there's no guarantee querying that block will be a good thing at all)

Done.

PR #140, opened 2020-07-17
Add support for TSDB on AWS

amckinley:
Currently, setting `storage_engine` to 'tsdb' assumes the use of GCS. Make it possible to use AWS instead.

### DIFF
--- a/cortex/config.libsonnet
+++ b/cortex/config.libsonnet
@@ -5,7 +5,7 @@
     replication_factor: 3,
     external_url: error 'must define external url for cluster',
 
-    storage_backend: error 'must specify storage backend (cassandra, gcp)',
+    storage_backend: error 'must specify storage backend (cassandra, gcp, aws)',
     table_prefix: $._config.namespace,
     cassandra_addresses: error 'must specify cassandra addresses',
     bigtable_instance: error 'must specify bigtable instance',
@@ -58,7 +58,7 @@
     // Use the Cortex chunks storage engine by default, while giving the ability
     // to switch to tsdb storage.
     storage_engine: 'chunks',
-    storage_tsdb_bucket_name: error 'must specify GCS bucket name to store TSDB blocks',
+    storage_tsdb_bucket_name: error 'must specify bucket name to store TSDB blocks',
 
     store_gateway_replication_factor: 3,
 
@@ -139,8 +139,6 @@
         'experimental.tsdb.block-ranges-period': '2h',
         'experimental.tsdb.retention-period': '96h',  // 4 days protection against blocks not being uploaded from ingesters.
         'experimental.tsdb.ship-interval': '1m',
-        'experimental.tsdb.backend': 'gcs',
-        'experimental.tsdb.gcs.bucket-name': $._config.storage_tsdb_bucket_name,
         'experimental.tsdb.store-gateway-enabled': true,
         'experimental.store-gateway.sharding-enabled': true,
         'experimental.store-gateway.sharding-ring.store': 'consul',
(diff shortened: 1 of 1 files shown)


amckinley:
This was originally https://github.com/grafana/cortex-jsonnet/pull/124, but I screwed up my repo in that PR.

cortex/config.libsonnet
```diff
@@ -139,15 +139,21 @@
...
         'experimental.store-gateway.sharding-ring.store': 'consul',
         'experimental.store-gateway.sharding-ring.consul.hostname': 'consul.%s.svc.cluster.local:8500' % $._config.namespace,
         'experimental.store-gateway.sharding-ring.prefix': '',
         'experimental.store-gateway.replication-factor': $._config.store_gateway_replication_factor,
       }
+    ) + (
+      if $._config.storage_backend == 'aws' then {
+        'experimental.tsdb.backend': 's3',
+        'experimental.tsdb.s3.bucket-name': $._config.storage_tsdb_bucket_name,
+      } else if $._config.storage_backend == 'gcs' then {
```
pracucci, line 153: As of today, `storage_backend` is used in the Cortex chunks storage and is a comma separated list of supported backends. Also, the expected value for GCS is `gcp`.

Since `storage_backend` is used to initialise `enabledBackends`, maybe a safer way to handle it is checking `if std.count($._config.enabledBackends, 'gcp') > 0`, similarly to what we do for `client_configs`. What do you think?

amckinley:
@pracucci how's it look now?

cortex/config.libsonnet
```diff
@@ -141,13 +143,22 @@
...
-        'experimental.tsdb.backend': 'gcs',
-        'experimental.tsdb.gcs.bucket-name': $._config.storage_tsdb_bucket_name,
+        'experimental.tsdb.store-gateway-enabled': true,
         'experimental.store-gateway.sharding-enabled': true,
         'experimental.store-gateway.sharding-ring.store': 'consul',
         'experimental.store-gateway.sharding-ring.consul.hostname': 'consul.%s.svc.cluster.local:8500' % $._config.namespace,
         'experimental.store-gateway.sharding-ring.prefix': '',
         'experimental.store-gateway.replication-factor': $._config.store_gateway_replication_factor,
+      }
+    ) + (
```
pracucci, line 153: Unfortunately we're not there yet. A couple of reasons:
1. This new config block you've added should be only applied when `$._config.storage_engine == 'tsdb' || $._config.querier_second_storage_engine == 'tsdb'` is `true` (see suggested change below)
2. Right now we're not setting `storage_backend` for the blocks storage. If you do (this PR) there're side effects introduced in the config originating by the fact that the `enabledBackends` is valued (eg. it will try to inject the bigtable config in `client_configs`)

The solution to 1 is something like this:
```
    storageTSDBConfig: (
      if $._config.storage_engine == 'tsdb' || $._config.querier_second_storage_engine == 'tsdb' then (
        {
          'store.engine': $._config.storage_engine,  // May still be chunks
          'experimental.tsdb.dir': '/data/tsdb',
          'experimental.tsdb.bucket-store.sync-dir': '/data/tsdb',
          'experimental.tsdb.bucket-store.ignore-deletion-marks-delay': '1h',
          'experimental.tsdb.block-ranges-period': '2h',
          'experimental.tsdb.retention-period': '96h',  // 4 days protection against blocks not being uploaded from ingesters.
          'experimental.tsdb.ship-interval': '1m',
          'experimental.tsdb.store-gateway-enabled': true,
          'experimental.store-gateway.sharding-enabled': true,
          'experimental.store-gateway.sharding-ring.store': 'consul',
          'experimental.store-gateway.sharding-ring.consul.hostname': 'consul.%s.svc.cluster.local:8500' % $._config.namespace,
          'experimental.store-gateway.sharding-ring.prefix': '',
          'experimental.store-gateway.replication-factor': $._config.store_gateway_replication_factor,
        } + (
          if std.count($._config.enabledBackends, 'aws') > 0 then {
            'experimental.tsdb.backend': 's3',
            'experimental.tsdb.s3.bucket-name': $._config.storage_tsdb_bucket_name,
            'experimental.tsdb.s3.endpoint': $._config.storage_tsdb_s3_endpoint,
          } else if std.count($._config.enabledBackends, 'gcp') > 0 then {
            'experimental.tsdb.backend': 'gcs',
            'experimental.tsdb.gcs.bucket-name': $._config.storage_tsdb_bucket_name,
          } else {}
        )
      )
      else {}
    ),
```

I believe the safest solution for 2 would be not using `storage_backend` and introduce a new config option `storage_tsdb_backend` which accept values `s3` and `gcs` (defaults to `gcs` so that we don't introduce a breaking change).

What do you think?

forestsword:
@amckinley are you interested in moving this forward? I am.

tomwilkie:
Very much so! Marco left a comment that needs addressing AFAICT, but we very much want to see this merged ASAP.  Let me know how I can help.

amckinley:
I'll try and get this cleaned up today.

PR #144, opened 2020-07-21
Improved blocks storage ingester alerts

pracucci:
In this PR I've improved the blocks storage ingester alerts:

- Do not fire `CortexIngesterHasNotShippedBlocks` or `CortexIngesterHasNotShippedBlocksSinceStart` if the ingester is not receiving any write traffic
- Added `CortexIngesterTSDBHeadCompactionFailed` to alert if TSDB head compaction is failing on a ingester (compaction is retried every minute)

### DIFF
--- a/cortex-mixin/alerts/blocks.libsonnet
+++ b/cortex-mixin/alerts/blocks.libsonnet
@@ -40,8 +40,9 @@
           },
         },
         {
-          // Alert if the ingester is failing to compact TSDB head into a block, for any opened TSDB. This is a critical
-          // condition that should never happen.
+          // Alert if the ingester is failing to compact TSDB head into a block, for any opened TSDB. Once the TSDB head is
+          // compactable, the ingester will try to compact it every 1 minute. Repeatedly failing it is a critical condition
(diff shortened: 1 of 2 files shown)


cortex-mixin/alerts/blocks.libsonnet
```diff
@@ -33,6 +39,21 @@
             message: 'Cortex Ingester {{ $labels.namespace }}/{{ $labels.instance }} has not shipped any block in the last 4 hours.',
           },
         },
+        {
+          // Alert if the ingester is failing to compact TSDB head into a block, for any opened TSDB. This is a critical
```
gotjosh, line 43: Could we add a note for how often we're supposed to compact the head? I don't remember (others could too) and is to hard to know how many opportunities we've had throughout the duration of the alert.

pracucci: Sure, makes sense. I've updated both. WDYT?

gotjosh: Looks good. Thanks!

PR #145, opened 2020-07-21
Added TSDB metrics to Cortex / Writes dashboard

pracucci:
In this PR I'm proposing to add more rows to the "Cortex / Writes" dashboard, adding metrics about TSDB.

Screen Shot 2020-07-21 at 14 51 06


codesome:
With alert I mean we would identify it quickly

pracucci:
> Would it make sense to add a panel for WAL corruption too?

Yes, a lot. Let me do it 🙏

pracucci:
> > Would it make sense to add a panel for WAL corruption too?
> 
> Yes, a lot. Let me do it 🙏

Done! I've updated the screenshot in the PR description.

PR #146, opened 2020-07-21
Add alert and dashboard for config file hashes

simonswine:
This allows to monitor the roll-out of new config file version to the various members of a cluster.

The metric itself was added as part of https://github.com/cortexproject/cortex/pull/2874.

Ideally someone double checks the filters I am using. I am also not 100% sure if really all nodes of a cluster typically have the exact same config and if the period of 1hours is reasonable.

This doesn't alert on runtime config divergence

scrn-2020-07-21-17-11-55
scrn-2020-07-21-17-11-45

### DIFF
--- a/cortex-mixin/alerts/alerts.libsonnet
+++ b/cortex-mixin/alerts/alerts.libsonnet
@@ -92,6 +92,21 @@
           },
         },
         {
+          alert: 'CortexInconsistentConfig',
+          expr: |||
+            count(count by(%s, job, sha256) (cortex_config_hash)) without(sha256) > 1
+          ||| % $._config.alert_aggregation_labels,
+          'for': '1h',
+          labels: {
+            severity: 'warning',
+          },
+          annotations: {
+            message: |||
+              An inconsistent config file hash is used across cluster {{ $labels.job }}.
+            |||,
+          },
+        },
+        {
           // As of https://github.com/cortexproject/cortex/pull/2092, this metric is
           // only exposed when it is supposed to be non-zero, so we don't need to do
           // any special filtering on the job label.
--- a/cortex-mixin/dashboards/config.libsonnet
+++ b/cortex-mixin/dashboards/config.libsonnet
@@ -1,0 +1,26 @@
+local utils = import 'mixin-utils/utils.libsonnet';
+
+(import 'dashboard-utils.libsonnet') {
+
+  'cortex-config.json':
+    $.dashboard('Cortex / Config')
+    .addClusterSelectorTemplates()
+    .addRow(
+      $.row('Startup config file')
+      .addPanel(
+        $.panel('Startup config file hashes') +
+        $.queryPanel('count(cortex_config_hash{%s}) by (sha256)' % $.namespaceMatcher(), 'sha256:{{sha256}}') +
+        $.stack +
+        { yaxes: $.yaxes('instances') },
+      )
+    )
+    .addRow(
+      $.row('Runtime config file')
+      .addPanel(
+        $.panel('Runtime config file hashes') +
+        $.queryPanel('count(cortex_runtime_config_hash{%s}) by (sha256)' % $.namespaceMatcher(), 'sha256:{{sha256}}') +
+        $.stack +
+        { yaxes: $.yaxes('instances') },
+      )
+    ),
+}
--- a/cortex-mixin/dashboards.libsonnet
+++ b/cortex-mixin/dashboards.libsonnet
@@ -1,5 +1,6 @@
 {
   grafanaDashboards+:
+    (import 'dashboards/config.libsonnet') +
     (import 'dashboards/queries.libsonnet') +
     (import 'dashboards/reads.libsonnet') +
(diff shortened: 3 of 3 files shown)


simonswine:
I think so, as the dashboard can be filtered by cluster and namespace through `.addClusterSelectorTemplates()` and the alert is using the `$._config.alert_aggregation_labels`.

I have tested a similar query (for the alert) with `cortex_build_info` as the hash metric is not yet deployed anywhere

cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -91,6 +91,21 @@
...
+          expr: |||
+            count(count by(%s, sha256) (cortex_config_hash)) without(sha256) > 1
+          ||| % $._config.alert_aggregation_labels,
+          'for': '1h',
+          labels: {
+            severity: 'warning',
+          },
+          annotations: {
+            message: |||
+              An inconsistent config file hash is used across cluster {{ $labels.job }}.
```
pracucci, line 105: `job` may not be among `alert_aggregation_labels` (it's not by default), so will be not valued here.

cortex-mixin/dashboards/config.libsonnet
```diff
@@ -0,0 +1,26 @@
...
+(import 'dashboard-utils.libsonnet') {
+
+  'cortex-config.json':
+    $.dashboard('Cortex / Config')
+    .addClusterSelectorTemplates()
+    .addRow(
+      $.row('Startup config file')
+      .addPanel(
+        $.panel('Startup config file hashes') +
+        $.queryPanel('count(cortex_config_hash) by (sha256)', 'sha256:{{sha256}}') +
```
pracucci, line 12: You should add the cluster selector label matchers (from the template), otherwise it counts it across all Cortex cluster and not for a specific one.

_Same comment applies to the panel below._

cortex-mixin/alerts/alerts.libsonnet
simonswine: I have added job to the by now, this should report the job back alright now

cortex-mixin/dashboards/config.libsonnet
simonswine: In the latest commit I am using `$.namespaceMatcher()`

cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -91,6 +91,21 @@
             |||,
           },
         },
+        {
+          alert: 'CortexInconsistentConfig',
+          expr: |||
+            count(count by(%s, job, sha256) (cortex_config_hash)) without(sha256) > 1
```
pracucci, line 97: Isn't this always > 1 when there are multiple jobs?

simonswine: I don't think so (maybe I am making a mistake here). 

I think the first `count` checks for series with differenent hashes and the second `count` filters out the ones from different clusters. This unit test passes (maybe you can break it :slightly_smiling_face: ):

```
jsonnet -S alerts.jsonnet > alerts.yaml

cat > test.yaml <<EOF  
rule_files:
  - alerts.yaml
evaluation_interval: 1m
tests:
 - interval: 1m
   input_series:
    - series: 'cortex_config_hash{sha256="aa",job="customer-a/cortex", cluster="customer-a", namespace="customer-a", instance="host-a1"}'
      values: '1+0x100'
    - series: 'cortex_config_hash{sha256="bb",job="customer-a/cortex", cluster="customer-a", namespace="customer-a", instance="host-a2"}'
      values: '1+0x100'
    - series: 'cortex_config_hash{sha256="cc",job="customer-b/cortex", cluster="customer-b", namespace="customer-b", instance="host-b1"}'
      values: '1+0x100'
    - series: 'cortex_config_hash{sha256="cc",job="customer-b/cortex", cluster="customer-b", namespace="customer-b", instance="host-b2"}'
      values: '1+0x100'
   alert_rule_test:
    - alertname: CortexInconsistentConfig
      eval_time: 59m
    - alertname: CortexInconsistentConfig
      eval_time: 60m
      exp_alerts:
       - exp_labels:
           severity: warning
           job: customer-a/cortex
           cluster: customer-a
           namespace: customer-a
         exp_annotations:
           message: "An inconsistent config file hash is used across cluster customer-a/cortex.\n"
EOF

promtool test rules unit_test.yaml
Unit Testing:  unit_test.yaml
  SUCCESS
```

jtlisi: I ran the following query against our local environment to verify the logic of this query:

```
count(count by(cluster, job, sha256) (cortex_runtime_config_hash)) without (sha256)
```

It came back looking sane to me

pracucci: I haven't had the time to get back to this. I'm sorry, but I'm not sure I do understand this alert. The query `count(...) without (sha256)` hides away different hashes, which is what we actually want to alert on (we want to alert when there are different hashes). What am I missing?

PR #147, opened 2020-07-24
Support Alertmanager HA

gotjosh:
With this, we can now support increasing the number of replicas for a
Cortex AM thus enabling HA.

 Please note that Alerts themselves are not gossiped between
Alertmanagers. Each Ruler needs to send the alert to every Alertmanager
available thus the reason why a headless service gets created when the
number of replicas is more than 1.

### DIFF
--- a/cortex/alertmanager.libsonnet
+++ b/cortex/alertmanager.libsonnet
@@ -4,8 +4,8 @@
   local container = $.core.v1.container,
   local statefulSet = $.apps.v1.statefulSet,
   local service = $.core.v1.service,
-  local isGossiping = $._config.alertmanager.replicas > 1,
-  local peers = if isGossiping then
+  local isHA = $._config.alertmanager.replicas > 1,
+  local peers = if isHA then
     [
       'alertmanager-%d.alertmanager.%s.svc.%s.local:%s' % [i, $._config.namespace, $._config.cluster, $._config.alertmanager.gossip_port]
       for i in std.range(0, $._config.alertmanager.replicas - 1)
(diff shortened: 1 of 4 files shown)


cortex/alertmanager.libsonnet
```diff
@@ -4,15 +4,23 @@
   local container = $.core.v1.container,
   local statefulSet = $.apps.v1.statefulSet,
   local service = $.core.v1.service,
+  local isGossiping = $._config.alertmanager.replicas > 1,
+  local peers = if isGossiping then
+    [
+      'alertmanager-%d.alertmanager.%s.svc.%s.local:%s' % [i, $._config.namespace, $._config.cluster, $._config.alertmanager.gossip_port]
+      for i in std.range(0, $._config.alertmanager.replicas - 1)
```
cstyan, line 11: this means a given alertmanager instance would have itself as a peer in it's cli arguments, could this be problematic?

cortex/alertmanager.libsonnet
```diff
@@ -50,6 +72,10 @@
 
   alertmanager_service:
     if $._config.alertmanager_enabled then
-      $.util.serviceFor($.alertmanager_statefulset)
+      if $._config.alertmanager.replicas > 1 then
```
cstyan, line 75: can we just use `isGossiping` here? I supposed the variable name could be confusing, since multiple AM instances doesn't necessarily mean we're gossiping, we're just treating it as such

maybe rename the var to `alertmanagerHA`?

cortex/alertmanager.libsonnet
gotjosh: Not from my experiments, the gossip protocol correctly identified that one of them was itself and whitelisted it.

cortex/alertmanager.libsonnet
gotjosh: Good shout.

cortex/config.libsonnet
```diff
@@ -227,6 +227,11 @@
         },
       }[$._config.ruler_client_type],
 
+    alertmanager: {
+      replicas: 1,
```
jtlisi, line 231: We are still going to default to 1 replica?

gouthamve: Yep, can we bump this to 3?

gouthamve (approved):

PR #148, opened 2020-07-27
Chunks blocks migration

pstibrany:
This PR adds some support for making chunks <-> blocks migration easier:

- adds ability to configure secondary store for querying
- modifies generation of blocks ingester's statefulset so that it's possible to instantiate it with different name.

### DIFF
--- a/cortex/ingester.libsonnet
+++ b/cortex/ingester.libsonnet
@@ -75,10 +75,12 @@
 
   local podDisruptionBudget = $.policy.v1beta1.podDisruptionBudget,
 
-  ingester_pdb:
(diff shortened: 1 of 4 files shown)


cortex/ingester.libsonnet
```diff
@@ -75,10 +75,12 @@
 
   local podDisruptionBudget = $.policy.v1beta1.podDisruptionBudget,
 
-  ingester_pdb:
+  newIngesterPdb(name, matchingNameLabel)::
```
pracucci, line 78: May be more clear if we rename `name > pdbName` and `matchingNameLabel > ingesterName`?

PR #150, opened 2020-07-29
Bump up limits for the ingester alerts

codesome:


cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -323,7 +323,7 @@
           alert: 'CortexProvisioningTooManyActiveSeries',
           // 1 million active series per ingester max.
```
pracucci, line 324: We should update this comment

cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -372,6 +372,23 @@
             |||,
           },
         },
+        {
+          alert: 'CortexProvisioningTooMuchMemory',
+          expr: |||
+            avg by (%s) (
+              container_memory_working_set_bytes{container_name="ingester"}
```
gouthamve, line 379: Can we not use average but, use something like %, using `container_spec_memory_limit_bytes`?

cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -372,6 +372,23 @@
             |||,
           },
         },
+        {
+          alert: 'CortexProvisioningTooMuchMemory',
```
pracucci, line 376: Why is it called "provisioning"? Maybe "Allocating" is more clear?

cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -372,6 +372,23 @@
...
+          expr: |||
+            avg by (%s) (
+              container_memory_working_set_bytes{container_name="ingester"}
+            ) > 11*1024*1024*1024
+          ||| % $._config.alert_aggregation_labels,
+          'for': '15m',
+          labels: {
+            severity: 'critical',
```
pracucci, line 384: Is it `critical` intentional? Doesn't necessarily means there's a critical issue, so it looks more a warning to me.

codesome: Right, this limit should be a warning and the 70% is critical.

codesome:
Updated the PR
- Comment now says 1.5M series (and made the alert for 1.6M for the series).
- Memory warning is now for 50% usage instead of hardcoded 11G.
- Alert name changed to `CortexAllocatingTooMuchMemory`

codesome:
Plus:
- The avg from the expression is removed

cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -354,14 +354,33 @@
...
+              container_spec_memory_limit_bytes{container_name="ingester"}
+            ) > 0.5
+          |||,
+          'for': '15m',
+          labels: {
+            severity: 'warning',
+          },
+          annotations: {
+            message: |||
+              Too much memory being used by ingesters - add more ingesters.
```
pracucci, line 371: Given now it's per ingester, we should mention it here using `{{ $labels.instance }}`. Same comment applies to the other alert.

cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -354,14 +354,33 @@
...
+          },
+        },
+        {
+          alert: 'CortexAllocatingTooMuchMemory',
+          expr: |||
+            (
               container_memory_working_set_bytes{container_name="ingester"}
                 /
               container_spec_memory_limit_bytes{container_name="ingester"}
             ) > 0.7
```
pracucci, line 382: 70% looks a bit low for a critical alert. This would have triggered over the last 7 days (try to run the query in our env). What's about 80%?

codesome: 80 sounds fine to me. @gouthamve?

gouthamve: I don't think so. 70% of the limit is too much for ingesters where we expect to be around 50%. I'd rather this pages quicker than later.

gouthamve: Actually nvm, I didn't realise the average was removed here also. This looks okay :)

PR #151, opened 2020-07-29
Update CortexAllocatingTooMuchMemory

codesome:
Added instance label and made the critical alert for 80% of limit

### DIFF
--- a/cortex-mixin/alerts/alerts.libsonnet
+++ b/cortex-mixin/alerts/alerts.libsonnet
@@ -368,7 +368,7 @@
           },
           annotations: {
             message: |||
-              Too much memory being used by ingesters - add more ingesters.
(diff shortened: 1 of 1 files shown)


cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -368,7 +368,7 @@
           },
           annotations: {
             message: |||
-              Too much memory being used by ingesters - add more ingesters.
+              Too much memory being used by {{ $labels.instance }}, - add more ingesters.
```
pracucci, line 371: `, -` looks too much 😉 

```suggestion
              Too much memory being used by {{ $labels.instance }} - add more ingesters.
```

codesome: Oops, a typo (I should stop hurrying :))

PR #154, opened 2020-08-07
Update blocks storage CLI flags

pracucci:
In https://github.com/cortexproject/cortex/pull/2937 we renamed some blocks storage CLI flags. This PR reflects such changes and requires you to run either Cortex `1.3.0-rc.0` or above, or `master`.

Breaking changes:
- `storage_engine` value changed from `tsdb` to `blocks`
- `querier_second_storage_engine` value changed from `tsdb` to `blocks`

### DIFF
--- a/cortex/tsdb.libsonnet
+++ b/cortex/tsdb.libsonnet
@@ -6,9 +6,9 @@
   local service = $.core.v1.service,
 
   _config+:: {
-    // Enforce TSDB storage
+    // Enforce blocks storage
     storage_backend: 'none',
-    storage_engine: 'tsdb',
+    storage_engine: 'blocks',
 
     // Allow to configure the ingester disk.
     cortex_ingester_data_disk_size: '100Gi',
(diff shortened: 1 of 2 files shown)


cortex/tsdb.libsonnet
```diff
@@ -96,7 +96,9 @@
     statefulSet.mixin.spec.template.metadata.withLabels({ name: name } + $.ingester_deployment_labels) +
     statefulSet.mixin.spec.selector.withMatchLabels({ name: name }) +
     statefulSet.mixin.spec.template.spec.securityContext.withRunAsUser(0) +
-    statefulSet.mixin.spec.template.spec.withTerminationGracePeriodSeconds(600) +
+    // When the ingester needs to flush blocks to the storage, it may take quite a lot of time.
```
gouthamve, line 99: hrm, this could potentially mean super slow rollout. Interesting.

pracucci: But we don't flush blocks on normal shutdown. We do only in special cases (eg. chunks to blocks migration).

PR #156, opened 2020-08-07
Fixed std.setMember() usage

pracucci:
Today I learned the hard way that `std.setMember()` requires the array to be sorted, while `std.member()` doesn't. So far it worked by coincidence, because we were defining `storage_engine: ['chunks', 'tsdb']` but after renaming `tsdb` to `blocks` it stopped working as expected. I replaced the usage of `std.setMember()` with `std.member()`.

To use the new function, I've upgraded the jsonnet requirement from `0.14.0` to `0.15.0`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,14 +10,14 @@ workflows:
 jobs:
   lint:
     docker:
-      - image: grafana/cortex-jsonnet-build-image:e6474c5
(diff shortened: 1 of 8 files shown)


pracucci:
_Investigating why the CI failed, while this change worked when I manually tried it._

pracucci:
`std.member()` is available since version 0.15.0.

PR #157, opened 2020-08-07
Add playbook entry for sample with repeated timestamp

MichelHollands:
This was added as a result of https://github.com/cortexproject/cortex/pull/2985. This documents reasons why the 'sample with repeated timestamp but different value' error appears in the log.

### DIFF
--- a/cortex-mixin/docs/playbooks.md
+++ b/cortex-mixin/docs/playbooks.md
@@ -1,6 +1,6 @@
 # Playbooks
 
-This document contains playbooks, or at least a checklist of what to look for, for alerts in the cortex-mixin. This document assumes that you are running a Cortex cluster:
+This document contains playbooks, or at least a checklist of what to look for, for alerts in the cortex-mixin and logs from Cortex. This document assumes that you are running a Cortex cluster:
 
 1. Using this mixin config
 2. Using GCS as object store (but similar procedures apply to other backends)
@@ -362,3 +362,16 @@ A PVC can be manually deleted by an operator. When a PVC claim is deleted, what 
 
 - `Retain`: the volume will not be deleted until the PV resource will be manually deleted from Kubernetes
 - `Delete`: the volume will be automatically deleted
+
+
+## Log lines
+
+### Log line containing 'sample with repeated timestamp but different value'
+
+This means a sample with the same timestamp as an existing one was received with a different value. The number of occurrences is recorded in the `prometheus_target_scrapes_sample_out_of_order_total` metric.
+
+Possible reasons for this are:
+- Multiple agents are scraping the same app without deduplication in place. Check the IP addresses mentioned in the log line for the agent that returned the deplicate sample. Change the labels of each sample generated per agent so they are unique.
+- Incorrect relabelling rules can cause a label to be dropped from a sample so that multiple samples have the same labels. If these samples were collected at the same time they will cause this error.
+- The exporter being scraped sets the same timestamp on every scrape. Note that exporters should generally not set timestamps.
+- Prometheus scrapes at the millisecond level. If the scrapes are done very quickly the same sample could be returned. This is very unlikely.
--- a/cortex-mixin/docs/playbooks.md
+++ b/cortex-mixin/docs/playbooks.md
@@ -368,7 +368,7 @@ A PVC can be manually deleted by an operator. When a PVC claim is deleted, what 
 
 ### Log line containing 'sample with repeated timestamp but different value'
 
-This means a sample with the same timestamp as an existing one was received with a different value. The number of occurrences is recorded in the `prometheus_target_scrapes_sample_out_of_order_total` metric.
+This means a sample with the same timestamp as the latest one was received with a different value. The number of occurrences is recorded in the `cortex_discarded_samples_total` metric with the label `reason="new-value-for-timestamp"`.
 
 Possible reasons for this are:
 - Multiple agents are scraping the same app without deduplication in place. Check the IP addresses mentioned in the log line for the agent that returned the duplicate sample. Change the labels of each sample generated per agent so they are unique.
--- a/cortex-mixin/docs/playbooks.md
+++ b/cortex-mixin/docs/playbooks.md
@@ -371,6 +371,5 @@ A PVC can be manually deleted by an operator. When a PVC claim is deleted, what 
 This means a sample with the same timestamp as the latest one was received with a different value. The number of occurrences is recorded in the `cortex_discarded_samples_total` metric with the label `reason="new-value-for-timestamp"`.
 
 Possible reasons for this are:
-- Multiple Prometheus servers / Grafana agents are scraping the same target without deduplication in place. Check the IP addresses mentioned in the log line for the agent that returned the duplicate sample. Change the labels of each sample generated per agent so they are unique.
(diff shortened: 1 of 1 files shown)


MichelHollands:
> _No description provided._

cortex-mixin/docs/playbooks.md
```diff
@@ -362,3 +362,16 @@ A PVC can be manually deleted by an operator. When a PVC claim is deleted, what
...
+
+
+## Log lines
+
+### Log line containing 'sample with repeated timestamp but different value'
+
+This means a sample with the same timestamp as an existing one was received with a different value. The number of occurrences is recorded in the `prometheus_target_scrapes_sample_out_of_order_total` metric.
+
+Possible reasons for this are:
+- Multiple agents are scraping the same app without deduplication in place. Check the IP addresses mentioned in the log line for the agent that returned the deplicate sample. Change the labels of each sample generated per agent so they are unique.
```
jdbaldry, line 374:
```suggestion
- Multiple agents are scraping the same app without deduplication in place. Check the IP addresses mentioned in the log line for the agent that returned the duplicate sample. Change the labels of each sample generated per agent so they are unique.
```

cortex-mixin/docs/playbooks.md
```diff
@@ -362,3 +362,16 @@ A PVC can be manually deleted by an operator. When a PVC claim is deleted, what
 
 - `Retain`: the volume will not be deleted until the PV resource will be manually deleted from Kubernetes
 - `Delete`: the volume will be automatically deleted
+
+
+## Log lines
+
+### Log line containing 'sample with repeated timestamp but different value'
+
+This means a sample with the same timestamp as an existing one was received with a different value. The number of occurrences is recorded in the `prometheus_target_scrapes_sample_out_of_order_total` metric.
```
pracucci, line 371:
```suggestion
This means a sample with the same timestamp as the latest one was received with a different value. The number of occurrences is recorded in the `cortex_discarded_samples_total` metric with the label `reason="new-value-for-timestamp"`.
```

cortex-mixin/docs/playbooks.md
```diff
@@ -362,3 +362,16 @@ A PVC can be manually deleted by an operator. When a PVC claim is deleted, what
...
+
+Possible reasons for this are:
+- Multiple agents are scraping the same app without deduplication in place. Check the IP addresses mentioned in the log line for the agent that returned the duplicate sample. Change the labels of each sample generated per agent so they are unique.
```
pracucci, line 374: More thoughts about it:
- Unlikely to be the cause. Timestamp are generally different when two independent processes scrape the same target. Timestamps could be the same when they're included in the `/metrics` response, but it's very uncommon (eg. among common exporters, I'm aware only cAdvisor doing it)
- Please always refer to Prometheus too as remote write source (it's more common than the Grafana agent)
- "Change the labels of each sample generated per agent" do you mean add an external label to the agent to make its series unique? Ok, but we shouldn't have multiple agents scraping the same targets unless they're configured as HA pairs in Cortex (see https://cortexmetrics.io/docs/production/ha-pair-handling/)

```suggestion
- Multiple Prometheus servers / Grafana agents are scraping the same target without deduplication in place. Check the IP addresses mentioned in the log line for the agent that returned the duplicate sample. Change the labels of each sample generated per agent so they are unique.
```

cortex-mixin/docs/playbooks.md
```diff
@@ -362,3 +362,16 @@ A PVC can be manually deleted by an operator. When a PVC claim is deleted, what
...
+- Incorrect relabelling rules can cause a label to be dropped from a sample so that multiple samples have the same labels. If these samples were collected at the same time they will cause this error.
```
pracucci, line 375: > If these samples were collected at the same time they will cause this error.

In other terms, this is generally caused when the clashing series are coming from the same exact target. So let's say a target expose these metrics:

```
container_cpu_usage_seconds{cpu="0"} 123
container_cpu_usage_seconds{cpu="1"} 456
```

Then you have a relabelling rule removing the cpu and you end up with clashing series with the same exact timestamp (because both series are scraped within the same scraping activity):
```
container_cpu_usage_seconds 123
container_cpu_usage_seconds 456
```

cortex-mixin/docs/playbooks.md
```diff
@@ -362,3 +362,16 @@ A PVC can be manually deleted by an operator. When a PVC claim is deleted, what
...
+- The exporter being scraped sets the same timestamp on every scrape. Note that exporters should generally not set timestamps.
+- Prometheus scrapes at the millisecond level. If the scrapes are done very quickly the same sample could be returned. This is very unlikely.
```
pracucci, line 377: I would remove this. Not a real case.

cortex-mixin/docs/playbooks.md
```diff
@@ -362,3 +362,15 @@ A PVC can be manually deleted by an operator. When a PVC claim is deleted, what
...
+
+## Log lines
+
+### Log line containing 'sample with repeated timestamp but different value'
+
+This means a sample with the same timestamp as the latest one was received with a different value. The number of occurrences is recorded in the `cortex_discarded_samples_total` metric with the label `reason="new-value-for-timestamp"`.
+
+Possible reasons for this are:
+- Multiple Prometheus servers / Grafana agents are scraping the same target without deduplication in place. Check the IP addresses mentioned in the log line for the agent that returned the duplicate sample. Change the labels of each sample generated per agent so they are unique.
+- Incorrect relabelling rules can cause a label to be dropped from a sample so that multiple samples have the same labels. If these samples were collected from the same target they will have the same timestamp. An example is dropping the `cpu` label when there are multiple cpus.
```
pracucci, line 375:
```suggestion
- Incorrect relabelling rules can cause a label to be dropped from a series so that multiple series have the same labels. If these series were collected from the same target they will have the same timestamp.
```

cortex-mixin/docs/playbooks.md
```diff
@@ -362,3 +362,15 @@ A PVC can be manually deleted by an operator. When a PVC claim is deleted, what
...
+
+
+## Log lines
+
+### Log line containing 'sample with repeated timestamp but different value'
+
+This means a sample with the same timestamp as the latest one was received with a different value. The number of occurrences is recorded in the `cortex_discarded_samples_total` metric with the label `reason="new-value-for-timestamp"`.
+
+Possible reasons for this are:
+- Multiple Prometheus servers / Grafana agents are scraping the same target without deduplication in place. Check the IP addresses mentioned in the log line for the agent that returned the duplicate sample. Change the labels of each sample generated per agent so they are unique.
```
pracucci, line 374: Few things, please:
1. As I mentioned in my previous comment https://github.com/grafana/cortex-jsonnet/pull/157#discussion_r467106893 I'm unsure this could be the cause for this specific error. Timestamps are different when the same target is monitored by different Prometheus servers. The only case could happen is when the monitored target expose metrics with timestamps **and** these timestamps don't change between the scraping done by two different Prometheus servers (eg. may happen with cAdvisor metrics).
2. Whenever you refer to "agent" only you should always mention "Prometheus / Agent" because the Agent is a less common case than running Prometheus and this is a public playbook

MichelHollands: Yeah, you're right. I spoke with Goutham about this specific problem which one customer was having. They had HA pairs setup that were not correctly using deduplication. But looking at my notes that might have been another issue. I'll remove this reason.

PR #158, opened 2020-08-10
Per Instance Read/Write Panels

joe-elliott:
This PR is designed to expose per instance metrics in the read/write dashboards to help diagnose issues quickly that turn out to be occurring due to the underlying infrastructure.  It is not uncommon during an incident to start writing queries that expose per pod metrics directly and to discover that the issues are due to 1 or very few pods.  These changes will allow operators to quickly identify an issue as being isolated to a small subset of pods and take appropriate measures.

- Adds panels to the read/write dashboards to see per instance metrics
  - requires a configuration option that defaults to empty
- Adds utility methods to accomplish the above
- Per pod metrics are purposefully only shown after a cluster and namespace have been picked.

image

### DIFF
--- a/cortex-mixin/config.libsonnet
+++ b/cortex-mixin/config.libsonnet
@@ -42,5 +42,8 @@
 
     // Whether resources dashboards are enabled (based on cAdvisor metrics).
     resources_dashboards_enabled: false,
+
+    // If supplied adds additional panels that are broken down per instance.  i.e. 'pod' in a kubernetes install
+    per_instance_label: '',
   },
 }
--- a/cortex-mixin/config.libsonnet
+++ b/cortex-mixin/config.libsonnet
@@ -44,6 +44,6 @@
     resources_dashboards_enabled: false,
 
     // If supplied adds additional panels that are broken down per instance.  i.e. 'pod' in a kubernetes install
-    per_instance_label: 'pod',
+    per_instance_label: '',
   },
 }
--- a/cortex-mixin/config.libsonnet
+++ b/cortex-mixin/config.libsonnet
@@ -43,7 +43,7 @@
     // Whether resources dashboards are enabled (based on cAdvisor metrics).
     resources_dashboards_enabled: false,
 
-    // If supplied adds additional panels that are broken down per instance.  i.e. 'pod' in a kubernetes install
-    per_instance_label: '',
+    // Used on panels that show metrics per instance.  i.e. 'pod' in a kubernetes install
+    per_instance_label: 'pod',
   },
 }
--- a/cortex-mixin/config.libsonnet
+++ b/cortex-mixin/config.libsonnet
@@ -44,6 +44,6 @@
     resources_dashboards_enabled: false,
 
     // If supplied adds additional panels that are broken down per instance.  i.e. 'pod' in a kubernetes install
-    per_instance_label: '',
+    per_instance_label: 'pod',
   },
 }
--- a/cortex-mixin/dashboards/dashboard-utils.libsonnet
+++ b/cortex-mixin/dashboards/dashboard-utils.libsonnet
@@ -78,10 +78,10 @@ local utils = import 'mixin-utils/utils.libsonnet';
       ],
     },
 
-  // hiddenLegendQueryPanel is a standard query panel designed to handle a large number of series.  it hides the legend, doesn't fill the series and 
+  // hiddenLegendQueryPanel is a standard query panel designed to handle a large number of series.  it hides the legend, doesn't fill the series and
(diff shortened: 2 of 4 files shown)


cortex-mixin/config.libsonnet
```diff
@@ -42,5 +42,8 @@
 
     // Whether resources dashboards are enabled (based on cAdvisor metrics).
     resources_dashboards_enabled: false,
+
+    // If supplied adds additional panels that are broken down per instance.  i.e. 'pod' in a kubernetes install
+    per_instance_label: 'pod',
```
tomwilkie, line 47: Will this work as "instance"?  Then it will work for the single binary case.

cortex-mixin/dashboards/reads.libsonnet
```diff
@@ -14,6 +14,14 @@ local utils = import 'mixin-utils/utils.libsonnet';
         $.panel('Latency') +
         utils.latencyRecordingRulePanel('cortex_request_duration_seconds', $.jobSelector($._config.job_names.gateway) + [utils.selector.re('route', 'api_prom_api_v1_.+')])
       )
+      .addPanelIf(
+        $._config.per_instance_label != '',
```
tomwilkie, line 18: Do we need this optionality?  Can we add these by default?

tomwilkie:
I think its fugly but I think its useful.  Can we make it more pretty?  Maybe this would be a good use for a topk? /cc @codesome...

cortex-mixin/dashboards/dashboard-utils.libsonnet
```diff
@@ -38,13 +38,26 @@ local utils = import 'mixin-utils/utils.libsonnet';
...
+    },
+
+  // The mixin allow specialism of the job selector depending on if its a single binary
   // deployment or a namespaced one.
   jobMatcher(job)::
     if $._config.singleBinary
     then 'job=~"$job"'
     else 'cluster=~"$cluster", job=~"($namespace)/%s"' % job,
 
+  jobMatcherEquality(job)::
```
pracucci, line 56: I'm not sure the function name matches what it does. To me it's like `jobMatcher()` but matching the namespace too.

cortex-mixin/config.libsonnet
joe-elliott: Yes, this will work as instance, but I meant to leave it blank to not show the new panels by default.  Removing.

cortex-mixin/dashboards/reads.libsonnet
joe-elliott: We could add them by default, but it requires a correct `per_instance_label` to work so I figured I'd only show the panels if the appropriate configuration was present.

Also unsure if everyone will see this as an upgrade or if some would prefer not showing it.

cortex-mixin/dashboards/dashboard-utils.libsonnet
joe-elliott: `jobMatcherEquality()` forces an exact match on both cluster and namespace.  This is to prevent drawing series on the new panels until both of these template options have been selected and only one of each is selected.   With "all" or multiple clusters selected those panels can get very busy. 

Another option might be to use the normal `jobMatcher()` and `topk` as suggested above to reduce drawn series, but I personally find `topk` graphs very confusing for dynamic series as it will often return an enormous number of always changing series.

I would be glad to go this route if desired.

cortex-mixin/dashboards/reads.libsonnet
tomwilkie: I'm concerned an explosion of optionality would make this library unmaintainable - I already had to go through once and simplify this.  I'd be very keen to make this non-optional.  Instance should work in almost all cases, and we can leave the label configurable for when it doesn't.

cortex-mixin/dashboards/dashboard-utils.libsonnet
pracucci: > This is to prevent drawing series on the new panels until both of these template options have been selected and only one of each is selected.

Now I understand it, thanks! I would add a comment to explain it :)

> I would be glad to go this route if desired.

I'm not sure `topk()` will be a better option because of the reason you described.

cortex-mixin/dashboards/reads.libsonnet
joe-elliott: Sounds good.  I will default to "pod" as it is a commonly used label in k8s series.

joe-elliott: Configurability removed.  Resolving.

PR #164, opened 2020-08-21
Release 1.3.0

### LINKED ISSUE #159: Changelog and versioning
This project is not versioned and has no changelog. For a non Grafana user it's quite difficult to understand which `cortex-jsonnet` version should run for a specific Cortex version. For example, the jsonnet `master` doesn't necessarily run with the latest Cortex stable version, because of config breaking changes done in the meanwhile for experimental features (eg. blocks storage).

I would **propose** to:
1. Introduce a `CHANGELOG.md` for any user-facing change
2. Cut a new version of the mixin every time a stable Cortex version is cut and keep the same versioning number (eg. `cortex-jsonnet` `1.3.0` should be used to run Cortex `1.3.0`)

Thoughts?

pracucci:
Today I released Cortex 1.3.0 and, as described in the issue #159, I would like to start versioning the jsonnet with the same Cortex version.

Fixes #159

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,7 @@
+**What this PR does**:
+
+**Which issue(s) this PR fixes**:
+Fixes #<issue number>
+
+**Checklist**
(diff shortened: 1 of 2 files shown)


pstibrany (approved):

pstibrany:
We forgot to bump Cortex image to 1.3 in this PR :)

PR #165, opened 2020-08-25
Fix annotation for CortexAllocatingTooMuchMemory

codesome:
Turns out `$labels.instance` is the node name in `container_memory_working_set_bytes`

### DIFF
--- a/cortex-mixin/alerts/alerts.libsonnet
+++ b/cortex-mixin/alerts/alerts.libsonnet
@@ -383,7 +383,7 @@
           },
           annotations: {
             message: |||
-              Too much memory being used by {{ $labels.pod }} - add more ingesters.
+              Too much memory being used by {{ $labels.namespace }}/{{ $labels.pod }} - add more ingesters.
             |||,
           },
         },
@@ -402,7 +402,7 @@
           },
           annotations: {
             message: |||
-              Too much memory being used by {{ $labels.pod }} - add more ingesters.
+              Too much memory being used by {{ $labels.namespace }}/{{ $labels.pod }} - add more ingesters.
             |||,
           },
         },
--- a/cortex-mixin/alerts/alerts.libsonnet
+++ b/cortex-mixin/alerts/alerts.libsonnet
@@ -383,7 +383,7 @@
           },
           annotations: {
             message: |||
-              Too much memory being used by {{ $labels.instance }} - add more ingesters.
+              Too much memory being used by {{ $labels.pod }} - add more ingesters.
             |||,
           },
         },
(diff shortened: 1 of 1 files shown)


pracucci (commented):
~~I'm not sure about this change because of few reasons:~~
1. ~~Alerts should be consistent (`$labels.pod` has been introduced now)~~
2. ~~I think we picked the path to not assume Cortex is deployed on Kubernetes~~

_Ignore this message. Was totally out of topic._

cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -383,7 +383,7 @@
           },
           annotations: {
             message: |||
-              Too much memory being used by {{ $labels.instance }} - add more ingesters.
+              Too much memory being used by {{ $labels.pod }} - add more ingesters.
```
pracucci, line 386:
```suggestion
              Too much memory being used by {{ $labels.namespace }}/{{ $labels.pod }} - add more ingesters.
```

cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -402,7 +402,7 @@
...
```
pracucci, line 405:
```suggestion
              Too much memory being used by {{ $labels.namespace }}/{{ $labels.pod }} - add more ingesters.
```

pracucci (approved):
LGTM (with a nit)

sandeepsukhani:
> ~I'm not sure about this change because of few reasons:~
> 
> 1. ~Alerts should be consistent (`$labels.pod` has been introduced now)~
> 2. ~I think we picked the path to not assume Cortex is deployed on Kubernetes~
> 
> _Ignore this message. Was totally out of topic, given these two alerts **are** intentionally based on container metrics._

@pracucci I was going to take over this PR and take care of deploying it internally.
I was wondering if we should replace all the usages of `$labels.instance` with `$labels.pod`. I haven't dug much deeper into it but I was thinking the `pod` label should be there for all the metrics most likely not just for container metrics, what do you think?

pracucci:
> I haven't dug much deeper into it but I was thinking the pod label should be there for all the metrics most likely not just for container metrics, what do you think?

@sandeepsukhani I don't think so. I believe we need to continue using "instance" to keep compatibility with deployments outside K8S, unless the metric used is K8S/container specific.

sandeepsukhani:
> > I haven't dug much deeper into it but I was thinking the pod label should be there for all the metrics most likely not just for container metrics, what do you think?
> 
> @sandeepsukhani I don't think so. I believe we need to continue using "instance" to keep compatibility with deployments outside K8S, unless the metric used is K8S/container specific.

So it means `job` label is added by k8s and is not related to the scrape config.
Sounds good. Will keep it as is and just take care of nits.

PR #169, opened 2020-08-26
exposes ruler queries for reuse

owen-d:
This extracts the ruler dashboard queries into a hidden field which can then be referenced. This is attractive for Loki in particular which wants to reuse the queries to monitor its ruler (built upon cortex). This is a noop.


gotjosh (approved):
As they are, these dashboards are not really all that useful 😭 .

It's on my agenda to create more graphs that help us spot iffy things more quickly - but it's currently at the bottom of my priorities. Let me know if you plan to do more work here and I'll be happy to run you through what I'm thinking.

PR #172, opened 2020-09-01
Run AlertManager as a non-root user.

mplzik:
**What this PR does**:

AlertManager has been a non-root docker image for some time already, so
this should be a no-op change (see https://github.com/prometheus/alertmanager/issues/1585)

Not reported as a bug.

**Checklist**
- [X] `CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] AlertManager libsonnet template uses a non-root user.
 * [CHANGE] The project is now licensed with Apache-2.0 license. #169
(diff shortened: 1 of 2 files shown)


gotjosh:
This is the Cortex AM instead of the Prometheus AM - Am I missing something or the same applies here?

mplzik:
@gotjosh true, I silently assumed that the AM image is the same as Prometheus one. If that's not the case, feel free to close this PR.

gotjosh:
@mplzik no worries, thanks for the pointer anyway!

PR #179, opened 2020-09-04
Fix typo in Alertmanager configuration and more tweaks

gotjosh:
This PR introduces three main things for our Alertmanager configuration:

- A small typo fix in the Alertmanager configuration `alertmanger.gossip_port` instead of `alertmanager_gossip_port`
- A new configuration options `fallback_config` to allow a fallback configuration to be specified.
- Uses a different name for the `headless-service`. The `ClusterIP` metadata field is immutable so we can't really update the spec of the service which makes it hard to switch between 1 and multiple replicas. 

**Checklist**

### DIFF
--- a/cortex/alertmanager.libsonnet
+++ b/cortex/alertmanager.libsonnet
@@ -47,7 +47,7 @@
       container.withArgsMixin(
         $.util.mapToFlags($.alertmanager_args) +
         if isHA then
-          ['--cluster.listen-address=[$(POD_IP)]:%s' % $._config.alertmanager_gossip_port] +
+          ['--cluster.listen-address=[$(POD_IP)]:%s' % $._config.alertmanager.gossip_port] +
           ['--cluster.peer=%s' % peer for peer in peers]
         else [],
       ) +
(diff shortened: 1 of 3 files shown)


jtlisi (approved):
LGTM

cortex/alertmanager.libsonnet
```diff
@@ -47,11 +62,16 @@
       container.withArgsMixin(
         $.util.mapToFlags($.alertmanager_args) +
         if isHA then
-          ['--cluster.listen-address=[$(POD_IP)]:%s' % $._config.alertmanager_gossip_port] +
+          ['--cluster.listen-address=[$(POD_IP)]:%s' % $._config.alertmanager.gossip_port] +
```
rfratto, line 65: Should this have a corresponding change somewhere or was it wrong before?

rfratto (approved):
LGTM, just one question

cortex/alertmanager.libsonnet
gotjosh: It was wrong all along 😭 , I introduced it as part of #147 which was merged a few hours ago. I state it in the body of the PR but didn't include it in the changelog given it wasn't part of a release.

rfratto: I need to get better at reading the PR description :) sorry for making you repeat yourself. LGTM!

gotjosh: Being fair, it was empty when I originally opened the PR 😆

PR #181, opened 2020-09-16
Block Storage Support for S3/AWS

### LINKED ISSUE #166: S3 Block Storage Support Incomplete
Hi everyone! I'd like to be able to use the following config and get block storage using aws s3:

```
local cortex = import 'cortex/cortex.libsonnet';

cortex {
  _config+:: {
    namespace: 'default',

    storage_backend: 'aws',
    storage_engine: 'blocks'
    aws_region: 'us-east-1',
    s3_bucket_name: 'my-bucket',
    ruler_client_type: 's3',

    blocksStorageConfig: 'have to override',
  },
}
```

But the `blocksStorageConfig` is currently hardcoded to gcs:

https://github.com/grafana/cortex-jsonnet/blob/76e5ce31383ab88aee7311e5267b1e2a5516f62b/cortex/config.libsonnet#L144-L145

I'd expect the configs above to configure block storage for s3 correctly.

forestsword:
Adds the setting `blocks_storage_backend` to allow switching between gcs and s3 for block storage. Feel free to critique the way I implemented it in jsonnet. Everyone has an opinion but I appreciate here that if I want to use azure I wouldn't need to add code here I could just add to `genericBlocksStorageConfig` and move on.

Closes #140, I tried to wait for it to complete. Fixes #166.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * [CHANGE] We now allow queries that are 32 days long. For example, rate(metric[32d]). Before it was 31d. #173
 * [ENHANCEMENT] Enable support for HA in the Cortex Alertmanager #147
 * [ENHANCEMENT] Support `alertmanager.fallback_config` option in the Alertmanager. #179
+* [ENHANCEMENT] Add support for S3 block storage. #181
 * [BUGFIX] Add support the `local` ruler client type  #175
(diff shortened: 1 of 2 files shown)


pracucci (approved):
LGTM, thanks! (and sorry for the delay)

PR #185, opened 2020-09-23
Lower default ingestion limits and create a new overrides user

gotjosh:
**What this PR does**:

Lowers the default limits for ingestion and makes the previous limits a new "user". 

**Checklist**

### DIFF
--- a/cortex/config.libsonnet
+++ b/cortex/config.libsonnet
@@ -281,14 +281,14 @@
       // ingester limits are 6s (#replicas) / 3x (#replication factor) higher.
       //
       // extra_small_user: {
-      //   ingestion_rate: 60,000
-      //   ingestion_burst_size: 600,000
+      //   ingestion_rate: 10,000
+      //   ingestion_burst_size: 200,000
       //
       //   max_series_per_user:   0 (disabled)
       //   max_series_per_metric: 0 (disabled)
       //
-      //   max_global_series_per_user:   600,000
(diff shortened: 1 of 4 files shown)


cortex/config.libsonnet
```diff
@@ -280,20 +280,34 @@
...
+      // extra_small_user: {
+      //   ingestion_rate: 60,000
+      //   ingestion_burst_size: 600,000
       //
       //   max_series_per_user:   0 (disabled)
       //   max_series_per_metric: 0 (disabled)
       //
-      //   max_global_series_per_user:   1,000,000
-      //   max_global_series_per_metric: 100,000
+      //   max_global_series_per_user:   600,000
```
gouthamve, line 290: Should be `100K` as this is global without replication, see: https://cortexmetrics.io/docs/configuration/arguments/#ingester-distributor--querier-limits

cortex/config.libsonnet
```diff
@@ -280,20 +280,34 @@
...
+      //   max_global_series_per_metric: 60,000
```
gouthamve, line 291: Make this 20K though, I can see users hitting against the 10K limit easily.

cortex/config.libsonnet
gotjosh: I knew we had that kind of documentation! Why doesn't it show up when you search for limits 😭

gouthamve (approved):

PR #188, opened 2020-09-26
Remove jobMatcherEquality; empty panels look worse than slow ones.

tomwilkie:
Eventually we'll use Ganesh's topk work in these panels.

**What this PR does**:

Remove jobMatcherEquality; this causes empty panels when you have multiple namespaces or clusters selected.  IMO this looks worse than slow ones.


gouthamve (approved):

joe-elliott (approved):
I'm fine with this change.  The original reason to require both cluster and namespace to match was that the results returned did not feel useful with multiple clusters/namespaces.  There were too many series displayed to have any value.

I think an ideal change may be to group by cluster/namespace if they don't match exactly (instead of showing per process metrics).  That way the panel would remain useful regardless of which options were selected. 

Either way, this is fine with me!

PR #189, opened 2020-09-27
feat: add new api endpoint prefixes to dashboards

jtlisi:
**What this PR does**:

- Include new HTTP endpoints as part of dashboard jsonnet


pracucci (approved):
Makes sense to me 👍

PR #192, opened 2020-09-30
Make sure only a single ruler rolls out at a time

gouthamve:
The ruler is part of the ring and I think should avoid large reshardings during rollouts.


pstibrany (approved):
Makes sense, to avoid overloading one ruler with too many rule groups. However since rulers generate new tokens, one can still expect distribution of the groups to be completely different at the end of deployment compared to the state before.

gotjosh (approved):

PR #199, opened 2020-10-11
feat: use querier specific API metrics

jtlisi:
**What this PR does**:

When running using a single binary the query-frontend API metrics will supersede the queriers for the same route. A separate set of latency metrics exist for the querier. This PR ensures those metrics are used by the Cortex read dashboard.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## master / unreleased
 
+* [ENHANCEMENT] Read dashboards now use `cortex_querier_request_duration_seconds` metrics to allow for accurate dashboards when deploying Cortex as a single-binary. #199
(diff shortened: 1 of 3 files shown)


simonswine (approved):
The queries and recording rules, LGTM.

I am not entirely sure where those inaccuracies are coming from in single binary mode. Does it relate to the fact that things like eventually auth are also part of the request duration in single binary mode?

jtlisi:
@simonswine In single-binary mode we don't register the querier handler using external HTTP routes. Those routes are registered for the query-frontend instead. The querier is registered on a secondary router that we pass requests to after they are processed by the query frontend and scheduled. This router has a separate set of latency metrics prefixed `cortex_querier_`.  The router for these metrics is used to in both single-binary and microservice mode and will have latency metrics for the querier in both cases.

pracucci (approved):

PR #200, opened 2020-10-12
Refactor limits and overrides

simonswine:
**What this PR does**:

Ensure we expose 'extra_small_user' and reference it setting the "default" values.

This will raise the limits of the 'small_user' preset to the defaults for `ingester.max-samples-per-query` and `ingester.max-series-per-query`. 

**Checklist**

### DIFF
--- a/cortex/ingester.libsonnet
+++ b/cortex/ingester.libsonnet
@@ -22,10 +22,12 @@
 
       // Limits config.
       'ingester.max-chunk-idle': $._config.max_chunk_idle,
-      'ingester.max-global-series-per-user': 150000,  // 150K
-      'ingester.max-global-series-per-metric': 20000,  // 20K
-      'ingester.max-series-per-user': 0,  // Disabled in favour of the max global limit
-      'ingester.max-series-per-metric': 0,  // Disabled in favour of the max global limit
+      'ingester.max-series-per-user': $._config.limits.max_series_per_user,
+      'ingester.max-series-per-metric': $._config.limits.max_series_per_metric,
+      'ingester.max-global-series-per-user': $._config.limits.max_global_series_per_user,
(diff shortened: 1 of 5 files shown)


cortex/ingester.libsonnet
```diff
@@ -22,10 +22,12 @@
...
-      'ingester.max-global-series-per-user': 150000,  // 150K
-      'ingester.max-global-series-per-metric': 20000,  // 20K
-      'ingester.max-series-per-user': 0,  // Disabled in favour of the max global limit
-      'ingester.max-series-per-metric': 0,  // Disabled in favour of the max global limit
+      'ingester.max-series-per-user': $._config.limits.max_series_per_user,
+      'ingester.max-series-per-metric': $._config.limits.max_series_per_metric,
+      'ingester.max-global-series-per-user': $._config.limits.max_global_series_per_user,
+      'ingester.max-global-series-per-metric': $._config.limits.max_global_series_per_metric,
+      'ingester.max-series-per-query': $._config.limits.max_series_per_query,
+      'ingester.max-samples-per-query': $._config.limits.max_samples_per_query,
```
simonswine, line 30: This will lower the default limits for `ingester.max-samples-per-query` from 1M to 100k and `ingester.max-series-per-query` from 100k to 10k.

We did not enforce this before as it looks like

gouthamve: I would keep it higher tbh. We overprovision our read path and I don't want tiny users to hit these limits and think that the queries are performant or scalable enough.

simonswine: Good point, this should be the "safest" path forward. Just updated the PR accordingly

gotjosh (approved):

PR #203, opened 2020-10-14
Add ingester-blocks to ingester's job label matcher

pracucci:
**What this PR does**:
When we migrate a Cortex cluster from chunks to blocks, the migration procedure involves creating a temporary statefulset named `ingester-blocks`. Because of this, during the migration we loose visibility over ingesters.

What's the sentiment adding `ingester-blocks` too to default job names?

N/A

**Checklist**

### DIFF
--- a/cortex-mixin/config.libsonnet
+++ b/cortex-mixin/config.libsonnet
@@ -26,7 +26,7 @@
     // These are used by the dashboards and allow for the simultaneous display of
     // microservice and single binary cortex clusters.
     job_names: {
-      ingester: '(ingester|ingester-blocks|cortex$)',  // Match also ingester-blocks, which is used during the migration from chunks to blocks.
+      ingester: '(ingester.*|cortex$)',  // Match also ingester-blocks, which is used during the migration from chunks to blocks.
       distributor: '(distributor|cortex$)',
       querier: '(querier|cortex$)',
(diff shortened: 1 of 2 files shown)


cortex-mixin/config.libsonnet
```diff
@@ -26,7 +26,7 @@
     // These are used by the dashboards and allow for the simultaneous display of
     // microservice and single binary cortex clusters.
     job_names: {
-      ingester: '(ingester|cortex$)',
+      ingester: '(ingester|ingester-blocks|cortex$)',  // Match also ingester-blocks, which is used during the migration from chunks to blocks.
```
pstibrany, line 29: Can we use `ingester.*`?

```suggestion
      ingester: '(ingester.*|cortex$)',  // Match also ingester-blocks, which is used during the migration from chunks to blocks.
```

pracucci: We could but we may include other custom jobs/deployments we don't want. Using `ingester-blocks` may be safer.

pstibrany (approved):

cortex-mixin/config.libsonnet
pstibrany: Why wouldn't we want all kind of ingesters?

jtlisi (approved):
LGTM

PR #205, opened 2020-10-17
Add the ruler to the read resources dashboard

gotjosh:
**What this PR does**:

Adds the ruler to the write resources dashboard. Technically the ruler is both a read/write resources but it felt more compelling to the add the write than the read. 

**Checklist**


cstyan (approved):

gouthamve:
I think its more read than write. The whole point of rules is to query large amounts of data and then distill it into small amounts of data. I would say it adds substantatively to read rather than write.

gotjosh:
good shout @gouthamve

PR #206, opened 2020-10-27
Added alert for failed compator run.

pstibrany:
**What this PR does**: This PR adds new alert for failed compactor run.

### DIFF
--- a/cortex-mixin/alerts/compactor.libsonnet
+++ b/cortex-mixin/alerts/compactor.libsonnet
@@ -63,6 +63,21 @@
             message: 'Cortex Compactor {{ $labels.namespace }}/{{ $labels.instance }} has not uploaded any block in the last 24 hours.',
           },
         },
+        {
+          // Alert if compactor fails.
+          alert: 'CortexCompactorRunFailed',
+          expr: |||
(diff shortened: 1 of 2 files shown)


cortex-mixin/alerts/compactor.libsonnet
```diff
@@ -63,6 +63,21 @@
...
           },
         },
+        {
+          // Alert if compactor fails.
+          alert: 'CortexCompactorRunFailed',
+          expr: |||
+            increase(cortex_compactor_runs_failed_total[10m]) > 0
+          |||,
+          labels: {
+            severity: 'warning',
```
pracucci, line 73: Warnings are generally ignored. Can we remove potential noise from the alert and make it a `critical` one?

For example, assuming there's some work to do at least every 2h, we could have a critical alert like this:
- expr: `increase(cortex_compactor_runs_failed_total{cluster="us-central1"}[3h]) > 0`
- for: `2h`

pstibrany: Updated based on offline discussion.

pracucci (approved):
LGTM, thanks!

PR #208, opened 2020-10-28
Added more critical alerts on Cortex ingester TSDB

pracucci:
**What this PR does**:
In the last few days we experienced TSDB WAL corruption twice and we found the alerts to not be adequate. In this PR I'm proposing to add more critical alerts in TSDB possible issue, including the usage of new metrics introduced in https://github.com/cortexproject/cortex/pull/3373.

N/A

**Checklist**

### DIFF
--- a/cortex-mixin/alerts/blocks.libsonnet
+++ b/cortex-mixin/alerts/blocks.libsonnet
@@ -56,6 +56,79 @@
           },
         },
         {
+          alert: 'CortexIngesterTSDBHeadTruncationFailed',
+          expr: |||
+            rate(cortex_ingester_tsdb_head_truncations_failed_total[5m]) > 0
+          |||,
+          labels: {
+            severity: 'critical',
+          },
+          annotations: {
+            message: 'Cortex Ingester {{ $labels.namespace }}/{{ $labels.instance }} is failing to truncate TSDB head.',
+          },
+        },
+        {
+          alert: 'CortexIngesterTSDBCheckpointCreationFailed',
+          expr: |||
+            rate(cortex_ingester_tsdb_checkpoint_creations_failed_total[5m]) > 0
+          |||,
+          labels: {
+            severity: 'critical',
+          },
+          annotations: {
+            message: 'Cortex Ingester {{ $labels.namespace }}/{{ $labels.instance }} is failing to create TSDB checkpoint.',
+          },
+        },
+        {
+          alert: 'CortexIngesterTSDBCheckpointDeletionFailed',
+          expr: |||
+            rate(cortex_ingester_tsdb_checkpoint_deletions_failed_total[5m]) > 0
+          |||,
+          labels: {
+            severity: 'critical',
+          },
+          annotations: {
+            message: 'Cortex Ingester {{ $labels.namespace }}/{{ $labels.instance }} is failing to delete TSDB checkpoint.',
+          },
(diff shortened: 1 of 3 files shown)


cortex-mixin/alerts/blocks.libsonnet
```diff
@@ -55,6 +55,79 @@
...
+            severity: 'critical',
+          },
+          annotations: {
+            message: 'Cortex Ingester {{ $labels.namespace }}/{{ $labels.instance }} is failing to truncate TSDB head.',
+          },
+        },
+        {
+          alert: 'CortexIngesterTSDBCheckpointCreationFailed',
+          expr: |||
+            rate(cortex_ingester_tsdb_checkpoint_creations_failed_total[5m]) > 0
```
codesome, line 73: Thinking out loud to make sure this is right:

For chunks WAL, we give a warning for 1 failure for checkpoint creation and critical for >1 failures in past 1h. This works because we checkpoint every 15m or so, hence we can capture >1 failures within 1h if there are back to back failures. 

With blocks, now the latest TSDB will only attempt checkpoint after 2h once it fails since the head truncation and checkpointing are separated. So immediately paging makes sense for 1 failure. Is there any case/requirement that we want to page on >1 failures? Hence check for >1 failures in last 3h?

cortex-mixin/alerts/blocks.libsonnet
```diff
@@ -55,6 +55,79 @@
...
+            message: 'Cortex Ingester {{ $labels.namespace }}/{{ $labels.instance }} is failing to delete TSDB checkpoint.',
+          },
+        },
+        {
+          alert: 'CortexIngesterTSDBWALTruncationFailed',
+          expr: |||
+            rate(cortex_ingester_tsdb_wal_truncations_failed_total[5m]) > 0
+          |||,
+          labels: {
+            severity: 'critical',
```
codesome, line 100: WAL truncation is just about deleting files. Do we want to be sensitive enough to page on a single failure?

cortex-mixin/alerts/blocks.libsonnet
codesome: I have seen some warnings in checkpoint creation failures for chunks storage and the failure seemed to be some transient issue with the disk which would recover soon. So maybe we want it to be less sensitive? (That also means we would miss the single failures if we are not looking at warning)

cortex-mixin/alerts/blocks.libsonnet
codesome: Because it could be a transient issue with disk again

codesome (commented):
(Realising after 5hrs that I had not clicked on submit :P)

cortex-mixin/alerts/blocks.libsonnet
pracucci: I would prefer to take a pessimistic approach and eventually fine tune the alert if generates false positive. I tried to run this query over the history of our Cortex blocks storage clusters and it would have fired only for issues which turned out to be incidents.

cortex-mixin/alerts/blocks.libsonnet
pracucci: It would have never triggered so far, but I agree with you we can lower it to warning.

codesome (approved):

PR #210, opened 2020-10-29
Removed support for ingester.statefulset_replicas

### LINKED ISSUE #209: `$._config.ingester.statefulset_replicas` not honoured for blocks storage ingesters
In https://github.com/grafana/cortex-jsonnet/blob/master/cortex/ingester.libsonnet#L92-L94 we can see: 

```
statefulSet.new('ingester', $._config.ingester.statefulset_replicas
``` 
So, the number of replicas comes from the config file value.

However, in https://github.com/grafana/cortex-jsonnet/blob/master/cortex/tsdb.libsonnet#L87-L88 we can see
```
newIngesterStatefulSet(name, container)::
    statefulSet.new(name, 3, [
```
So if using tsdb.libsonnet you get ingesters hardcoded to 3, which seems odd.

cc @pracucci as filing this per his request.

pracucci:
**What this PR does**:
The issue #209 make me realise the chunks storage with WAL enabled is using a pattern different than any other place in our jsonnet to configure the number of replicas. Instead of fixing #209 applying the config option to blocks storage too, I'm proposing to remove that config option at all, because it's looks like breaking our config consistency.

Fixes #209

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 ## master / unreleased
 
 * [CHANGE] Add the default preset 'extra_small_user' and reference it in the CLI flags. This will raise the limits of the 'small_user' preset to the defaults for `ingester.max-samples-per-query` and `ingester.max-series-per-query`. #200
+* [CHANGE] Removed the config option `$._config.ingester.statefulset_replicas` which was used only when running Cortex chunks storage with WAL enabled. To configure the number of ingester replicas you should now use the following: #210
+  ```
+  ingester_statefulset+:
+      statefulSet.mixin.spec.withReplicas(6),
+  ```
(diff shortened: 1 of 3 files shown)


tomwilkie (approved):

bboreham:
I tried adding the lines given in the changelog, but I get `Unknown variable: statefulSet`.
Could you give more context on how they should be used?

pracucci:
> I tried adding the lines given in the changelog, but I get `Unknown variable: statefulSet`.
> Could you give more context on how they should be used?

@bboreham Yes, sure. Something like this:

```
local statefulSet = $.apps.v1.statefulSet,

ingester_statefulset+:
    statefulSet.mixin.spec.withReplicas(9),
```

PR #214, opened 2020-11-10
Fix severity for CortexCompactorRunFailed alert.

pstibrany:
Per discussion in #206


pracucci (approved):
Dance, dance, dance 😄

PR #222, opened 2020-11-20
[fixup] ruler limits config key name

gotjosh:
**What this PR does**:

Ruler limits have a prefix of `ruler_` on the config key name. This
makes the key match and then uses them as the value for the flags.

**Checklist**

### DIFF
--- a/cortex/ruler.libsonnet
+++ b/cortex/ruler.libsonnet
@@ -20,8 +20,8 @@
       'ruler.ring.consul.hostname': 'consul.%s.svc.cluster.local:8500' % $._config.namespace,
 
       // Limits
-      'ruler.max-rules-per-rule-group': $._config.limits.max_rules_per_rule_group,
-      'ruler.max-rule-groups-per-tenant': $._config.limits.max_rule_groups_per_tenant,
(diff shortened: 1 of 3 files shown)


cortex/ruler.libsonnet
```diff
@@ -20,8 +20,8 @@
       'ruler.ring.consul.hostname': 'consul.%s.svc.cluster.local:8500' % $._config.namespace,
 
       // Limits
-      'ruler.max-rules-per-rule-group': $._config.limits.max_rules_per_rule_group,
-      'ruler.max-rule-groups-per-tenant': $._config.limits.max_rule_groups_per_tenant,
+      'ruler.max-rules-per-rule-group': $._config.limits.ruler_max_rules_per_rule_group,
```
pracucci, line 23: Where are defined the default values for these?

pracucci (approved):

cortex/ruler.libsonnet
gotjosh: Discussed this offline, we set a specific user as the defaults for the `limits` key.

PR #224, opened 2020-12-04
Uses $__rate_interval in ruler dashboard queries

owen-d:
Using the `$__interval` along with rate/increase queries can leave undesirably blank dashboards in some cases. As per https://grafana.com/docs/grafana/latest/datasources/prometheus/#using-__rate_interval-variable, we should use `$__rate_interval` instead.

Before: 
image

After:
image


cstyan (approved):

pracucci (approved):
LGTM, thanks! We're using `$__interval` with `rate()` in other dashboards too. Do you mind opening a PR for that too, please?

PR #227, opened 2020-12-11
Improve blocks storage observability

pracucci:
**What this PR does**:
In this PR I've introduced new panels and an alert related to the blocks storage, to improve the visibility we have over all operational aspects:

- Cortex / Writes
  - Headlines: added current number of tenants in the cluster
- Cortex / Writes Resources
  - Added ingester disk read/writes/utilisation
- Cortex / Reads Resources
  - Added store-gateway disk read/writes/utilisation
- Cortex / Queries
  - Added "Lazy loaded index-headers"
  - Added "Lazy load duration"
- Cortex / Compactor
  - Added "Tenants compaction progress"
- Alerts
  - CortexMemoryMapAreasTooHigh

_I manually tested the dashboards and should work as expected._

N/A

**Checklist**

### DIFF
--- a/cortex-mixin/alerts/alerts.libsonnet
+++ b/cortex-mixin/alerts/alerts.libsonnet
@@ -242,7 +242,7 @@
         {
           alert: 'CortexMemoryMapAreasTooHigh',
           expr: |||
-            process_memory_map_areas{job=~".+(cortex|ingester)"} / process_memory_map_areas_limit{job=~".+(cortex|ingester)"} > 0.8
+            process_memory_map_areas{job=~".+(cortex|ingester|store-gateway)"} / process_memory_map_areas_limit{job=~".+(cortex|ingester|store-gateway)"} > 0.8
           |||,
           'for': '5m',
           labels: {
--- a/cortex-mixin/alerts/alerts.libsonnet
+++ b/cortex-mixin/alerts/alerts.libsonnet
@@ -239,6 +239,19 @@
             |||,
           },
         },
+        {
+          alert: 'CortexMemoryMapAreasTooHigh',
+          expr: |||
+            process_memory_map_areas{job=~".+(cortex|ingester)"} / process_memory_map_areas_limit{job=~".+(cortex|ingester)"} > 0.8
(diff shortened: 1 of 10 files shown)


cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -224,6 +224,19 @@
             |||,
           },
         },
+        {
+          alert: 'CortexMemoryMapAreasTooHigh',
+          expr: |||
+            process_memory_map_areas{job=~".+(cortex|ingester)"} / process_memory_map_areas_limit{job=~".+(cortex|ingester)"} > 0.8
```
pstibrany, line 230: Worth including store-gateway too?

cortex-mixin/docs/playbooks.md
```diff
@@ -53,26 +53,17 @@ This alert goes off when an ingester fails to find another node to transfer its
...
+### CortexMemoryMapAreasTooHigh
 
-### CortexLoadBalancerErrors
-@todo
+This alert fires when a Cortex process has a number of memory map areas close to the limit. The limit is a per-process limit imposed by the kernel and this issue is typically caused by a large number of mmap-ed failes.
 
-### CortexTableSyncFailure
-@todo
+How to **fix**:
+- Increase the limit on your system: `sysctl -w vm.max_map_count=<NEW LIMIT>`
```
pstibrany, line 61: This needs to be performed on the node, right? It may not be trivial to change.

pstibrany (approved):
LGTM

cortex-mixin/alerts/alerts.libsonnet
pracucci: Definitely. It was intended to cover store-gateway too and was a mistake not including it.

cortex-mixin/docs/playbooks.md
pracucci: Yes, on the node. In our infra it's easy because we have a deamonset running `startup.sh` but playbook is generic, so "it depends".

PR #229, opened 2020-12-14
Fixed workingset memory panel while rolling out a StatefulSet

pracucci:
**What this PR does**:
Everytime we rollout ingesters, I see this worrying trend:
Screenshot 2020-12-14 at 10 01 26

Today I've investigated it and it turned out to be a false positive. The problem is that during the rollout of a statefulset we have duplicated series, both the old pod and the new pod (even if the pod ID is the same, the metric have other labels which are distinct).

In the following panel, on the left you can see `sum by(pod)` while on the right you can see the 2 series:
Screenshot 2020-12-14 at 10 01 37

This PR should fix it.

N/A

**Checklist**


pstibrany (approved):
Nice find!

PR #236, opened 2021-01-05
Add option to configure unregister ingesters on shutdown

pracucci:
**What this PR does**:
In this PR I've added an option to allow to configure the "unregister ingesters on shutdown" (enabled by default). The CLI flags are added conditionally in order to avoid any config change unless you disable the unregistering.

N/A

**Checklist**

### DIFF
--- a/cortex/config.libsonnet
+++ b/cortex/config.libsonnet
@@ -15,8 +15,9 @@
 
     // If false, ingesters are not unregistered on shutdown and left in the ring with
     // the LEAVING state. Setting to false prevents series resharding during ingesters rollouts,
-    // but requires to manually forget ingesters on scale down and that ingester ID is preserved
-    // during rollouts.
+    // but requires to:
+    // 1. Either manually forget ingesters on scale down or invoke the /shutdown endpoint
+    // 2. Ensure ingester ID is preserved during rollouts
     unregister_ingesters_on_shutdown: true,
 
     // schema is used to generate the storage schema yaml file used by
--- a/cortex/config.libsonnet
+++ b/cortex/config.libsonnet
@@ -13,8 +13,8 @@
     aws_region: error 'must specify AWS region',
     s3_bucket_name: error 'must specify S3 bucket name',
 
-    // If disabled, ingesters are not unregistered on shutdown and left in the ring with
-    // the LEAVING state. Disabling it prevents series resharding during ingesters rollouts,
+    // If false, ingesters are not unregistered on shutdown and left in the ring with
+    // the LEAVING state. Setting to false prevents series resharding during ingesters rollouts,
     // but requires to manually forget ingesters on scale down and that ingester ID is preserved
     // during rollouts.
     unregister_ingesters_on_shutdown: true,
--- a/cortex/config.libsonnet
+++ b/cortex/config.libsonnet
@@ -13,6 +13,12 @@
     aws_region: error 'must specify AWS region',
     s3_bucket_name: error 'must specify S3 bucket name',
 
+    // If disabled, ingesters are not unregistered on shutdown and left in the ring with
+    // the LEAVING state. Disabling it prevents series resharding during ingesters rollouts,
+    // but requires to manually forget ingesters on scale down and that ingester ID is preserved
+    // during rollouts.
+    unregister_ingesters_on_shutdown: true,
+
     // schema is used to generate the storage schema yaml file used by
     // the Cortex chunks storage:
     // - More information: https://github.com/cortexproject/cortex/pull/1072
--- a/cortex/distributor.libsonnet
+++ b/cortex/distributor.libsonnet
@@ -37,8 +37,8 @@
       'distributor.ring.consul.hostname': 'consul.%s.svc.cluster.local:8500' % $._config.namespace,
       'distributor.ring.prefix': '',
 
-      // Do not extend the replication set on unhealthy ingester when "unregister on shutdown"
-      // is disabled.
+      // Do not extend the replication set on unhealthy (or LEAVING) ingester when "unregister on shutdown"
+      // is set to false.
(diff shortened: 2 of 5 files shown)


cortex/config.libsonnet
```diff
@@ -13,6 +13,12 @@
     aws_region: error 'must specify AWS region',
     s3_bucket_name: error 'must specify S3 bucket name',
 
+    // If disabled, ingesters are not unregistered on shutdown and left in the ring with
```
pstibrany, line 16: "If disabled" -- what *exactly* does it mean here? Does "unregister: true" mean disabled or enabled?

I would suggest to be explicit: "If set to true/false...". (Also on next line "Disabling it...")

cortex/querier.libsonnet
```diff
@@ -27,7 +27,13 @@
...
 
       'log.level': 'debug',
-    },
+    } + (
+      if !$._config.unregister_ingesters_on_shutdown then
+        {
+          'distributor.extend-writes': false,
+        }
+      else {}
+    ),
```
pstibrany, line 36: Shouldn't this be part of distributor args, instead of querier?

pracucci: Uh ops. Yes definitely, was a too quick PR.

cortex/config.libsonnet
pracucci: You're definitely right. What's about now?

cortex/config.libsonnet
```diff
@@ -13,6 +13,12 @@
     aws_region: error 'must specify AWS region',
     s3_bucket_name: error 'must specify S3 bucket name',
 
+    // If false, ingesters are not unregistered on shutdown and left in the ring with
+    // the LEAVING state. Setting to false prevents series resharding during ingesters rollouts,
+    // but requires to manually forget ingesters on scale down and that ingester ID is preserved
```
pstibrany, line 18: Note that using `/shutdown` will still unregister, regardless of the setting. Should we mention it here?

cortex/distributor.libsonnet
```diff
@@ -36,7 +36,13 @@
...
       'distributor.ring.consul.hostname': 'consul.%s.svc.cluster.local:8500' % $._config.namespace,
       'distributor.ring.prefix': '',
-    },
+    } + (
+      if !$._config.unregister_ingesters_on_shutdown then
+        {
+          'distributor.extend-writes': false,
+        }
+      else {}
+    ),
```
pstibrany, line 45: We can simplify with:

```suggestion
          'distributor.extend-writes': $._config.unregister_ingesters_on_shutdown,
        }
```

As it defaults to true anyway.

cortex/ingester.libsonnet
```diff
@@ -39,6 +39,12 @@
           'store.index-cache-write.memcached.service': 'memcached-client',
         }
       else {}
+    ) + (
+      if !$._config.unregister_ingesters_on_shutdown then
+        {
+          'ingester.unregister-on-shutdown': false,
+        }
+      else {}
```
pstibrany, line 47: We can also use `'ingester.unregister-on-shutdown': $._config.unregister_ingesters_on_shutdown` directly here, without `if`.

cortex/config.libsonnet
pstibrany: That's easier to understand now, thanks!

cortex/ingester.libsonnet
pracucci: Yeah, but that will introduce a config change across all Cortex clusters as soon as you update the mixin. Should we do it?

pstibrany: Oh, that's a good point. Personally I'd weakly vote for "yes", but I'm less enthusiastic about it now, and understand if you prefer it this way.

pracucci: Let me ask a second opinion. /cc @gouthamve

gouthamve: I would vote for "yes" as well imo. Really dislike all the if statement complexity.

Having said that, this change should ideally go out in a release and then pushed here. That way our clusters will have no impact when we update. But then we should also think about how we would want communicate the change to other users depending on this.

pstibrany: We could also update mixin only in environments which are being updated, and only update it globally once it's everywhere. But that sounds like lot of pain.

cortex/config.libsonnet
pracucci: Sure, done.

cortex/ingester.libsonnet
pracucci: > this change should ideally go out in a release and then pushed here

Sounds complicated. I think we should aim to move fast without breaking stuff. If we don't like the `if`, then I can remove it and just merge it. As soon as we update the mixin we'll see the diff until the next rollout. That's life.

pstibrany: > As soon as we update the mixin we'll see the diff until the next rollout. That's life.

Flux will also redeploy environments that it manages.

pstibrany (approved):

cortex/distributor.libsonnet
```diff
@@ -36,7 +36,11 @@
       // The ingestion rate global limit requires the distributors to form a ring.
       'distributor.ring.consul.hostname': 'consul.%s.svc.cluster.local:8500' % $._config.namespace,
       'distributor.ring.prefix': '',
-    },
+
+      // Do not extend the replication set on unhealthy ingester when "unregister on shutdown"
+      // is disabled.
```
pstibrany, line 41:
```suggestion
      // Do not extend the replication set on unhealthy ingester (LEAVING) when "unregister on shutdown"
      // is set to false.
```

cortex/ingester.libsonnet
pracucci: Then let's merge it before we rollout flux everywhere :)

PR #239, opened 2021-01-05
Honor per_instance_label in all panels

pracucci:
**What this PR does**:
While working on other changes in the mixin, I've realised we're not honoring everywhere the configured `per_instance_label`, but we have hardcoded `instance` or `pod` in some panels. In this PR I'm fixing it, excluding the container metrics which are still using the `pod` label hardcoded.

_I manually tested all changes and should be good._

N/A

**Checklist**


pstibrany (approved):
> I manually tested all changes and should be good.

🤞 🙈

PR #242, opened 2021-01-06
Ignore long-running requests on query-scheduler when checking for high latency.

pstibrany:
**What this PR does**: This PR updates `CortexRequestLatency` to ignore long-running requests on query-scheduler.

**Checklist**

### DIFF
--- a/cortex-mixin/alerts/alerts.libsonnet
+++ b/cortex-mixin/alerts/alerts.libsonnet
@@ -39,7 +39,7 @@
         {
           alert: 'CortexRequestLatency',
           expr: |||
-            cluster_namespace_job_route:cortex_request_duration_seconds:99quantile{route!~"metrics|/frontend.Frontend/Process|ready|/schedulerpb.SchedulerForFrontend/FrontendLoop|/schedulerpb.SchedulerForFrontend/FrontendLoop"}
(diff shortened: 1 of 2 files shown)


cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -39,7 +39,7 @@
         {
           alert: 'CortexRequestLatency',
           expr: |||
-            cluster_namespace_job_route:cortex_request_duration_seconds:99quantile{route!~"metrics|/frontend.Frontend/Process|ready"}
+            cluster_namespace_job_route:cortex_request_duration_seconds:99quantile{route!~"metrics|/frontend.Frontend/Process|ready|/schedulerpb.SchedulerForFrontend/FrontendLoop|/schedulerpb.SchedulerForFrontend/FrontendLoop"}
```
gouthamve, line 42: Why is it repeated twice?

gouthamve (approved):
LGTM once the repetition is fixed

cortex-mixin/alerts/alerts.libsonnet
pstibrany: There are two different paths on query-scheduler (one for frontend and one for querier), but I included same route twice, instead of two routes. It's fixed now.

PR #251, opened 2021-01-19
Disable multi-selection for variables in resources dashboards.

pstibrany:
**What this PR does**: This PR disables multi-selection of `cluster` and `namespace` variable in resources dashboards.

I think we should do it in all dashboards, but this conservative PR is a first step to see what others think about it.

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [CHANGE] Only single cluster and namespace can now be selected in "resources" dashboards. #251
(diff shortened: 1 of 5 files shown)


pracucci (approved):
In resources dashboards make a lot of sense to me. Don't have a strong opinion about other dashboards tho.

pstibrany:
My reason is that I usually (99%) want to see dashboards for a single namespace, but the way Grafana works, I always need to do couple of clicks to get to the correct state.

PR #253, opened 2021-01-19
Fix -querier.second-store-engine in the ruler

pracucci:
**What this PR does**:
The ruler internally runs the querier, so any querier config option applies to ruler too. I just realised that, when `querier_second_storage_engine` is configured, we're not setting `-querier.second-store-engine` in the ruler too. This PR fixes it.

N/A

**Checklist**


gotjosh (approved):

gotjosh:
I don't see the actual fix though?

pstibrany (approved):
👍

gotjosh (approved):

PR #254, opened 2021-01-20
Added bucket index flag, and enable bucket index by default.

pstibrany:
**What this PR does**: This PR adds flag for bucket index, and enables it by default.

**Checklist**


cortex/tsdb.libsonnet
```diff
@@ -24,6 +24,12 @@
 
     // Allow to fine tune compactor.
     cortex_compactor_max_concurrency: 1,
+    // While this is the default value, we want to pass the same to the -blocks-storage.bucket-store.sync-interval
+    cortex_compactor_cleanup_interval: '15m',
+
+    // Enable use of bucket index by querier, ruler and store-gateway.
+    // Bucket index is generated by compactor from Cortex 1.7, there is no flag required to enable this on compactor.
+    cortex_bucket_index_enabled: true,
```
pracucci, line 32: I'm dubious about the default. I think default should be false because, when upgrading Cortex from 1.6 to 1.7, you need to have a full compactor run before enabling it.

pracucci (approved):
LGTM, just a comment about the default value

PR #257, opened 2021-02-05
Cleanup blocks storage config

pracucci:
**What this PR does**:
When we introduced the blocks store support to jsonnet we picked the easiest path: define a `genericBlocksStorageConfig` and set all CLI flags there. These flags were applied to all Cortex services accessing the  storage (ingester, querier, ruler, store-gateway, compactor).

However, not all flags we set in `genericBlocksStorageConfig` are used by all Cortex services. For example, the flag with prefix `-blocks-storage.tsdb.` are ingester specific.

In this PR I'm cleaning up the config, setting each CLI flag (related to blocks storage) only to Cortex services that actually use it.

N/A

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * [CHANGE] Only single cluster and namespace can now be selected in "resources" dashboards. #251
 * [CHANGE] Increased `CortexAllocatingTooMuchMemory` warning alert threshold from 50% to 65%. #256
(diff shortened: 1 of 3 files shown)


pstibrany (approved):
LGTM.

`$._config.queryBlocksStorageConfig`, `$.blocks_metadata_caching_config` and `$.bucket_index_config` seem to be always used together. Maybe the latter two could be merged into first one?

pracucci:
> $._config.queryBlocksStorageConfig, $.blocks_metadata_caching_config and $.bucket_index_config seem to be always used together. Maybe the latter two could be merged into first one?

Ok. Let me do it in a separate PR.

PR #258, opened 2021-02-08
feat: allow for Alertmanager to configure multiple storage backends

jtlisi:
**What this PR does**:

- Create a separate config object for the Alertmanager storage backend

**Checklist**

### DIFF
--- a/cortex/alertmanager.libsonnet
+++ b/cortex/alertmanager.libsonnet
@@ -18,7 +18,7 @@
 
   alertmanager_args::
     $._config.grpcConfig +
-    $._config.alertmanagerClientConfig +
+    $._config. alertmanagerStorageClientConfig +
     {
       target: 'alertmanager',
       'log.level': 'debug',
--- a/cortex/alertmanager.libsonnet
+++ b/cortex/alertmanager.libsonnet
@@ -18,14 +18,12 @@
 
   alertmanager_args::
     $._config.grpcConfig +
+    $._config.alertmanagerClientConfig +
     {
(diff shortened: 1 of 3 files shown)


mattmendick (approved):
LGTM

cortex/config.libsonnet
```diff
@@ -283,6 +283,29 @@
       fallback_config: {},
     },
 
+    alertmanager_client_type: error 'you must specify a storage backend type for the ruler (azure, configdb, gcs, s3, local)',
+    alertmanager_s3_bucket_name: $._config.s3_bucket_name,
+    alertmanager_gcs_bucket_name: error 'must specify a GCS bucket name',
+
+    alertmanagerClientConfig:
```
gotjosh, line 290:
```suggestion
    alertmanagerStorageClientConfig:
```

cortex/alertmanager.libsonnet
```diff
@@ -18,14 +18,12 @@
 
   alertmanager_args::
     $._config.grpcConfig +
+    $._config.alertmanagerClientConfig +
```
gotjosh, line 21:
```suggestion
    $._config. alertmanagerStorageClientConfig +
```

gotjosh (approved):
This is effectively a breaking change - could we state it as part of the

jtlisi:
> This is effectively a breaking change - I'm assuming we don't list those anywhere. right?

I added a entry to the `CHANGELOG.md` to signal that is a breaking change

PR #259, opened 2021-02-12
Enable @ modifier in PromQL

gouthamve:


cortex/querier.libsonnet
```diff
@@ -26,6 +26,9 @@
 
       'querier.second-store-engine': $._config.querier_second_storage_engine,
 
+      // Enable experimental @ support in PromQL.
+      'querier.at-modifier-enabled': true,
```
pracucci, line 30: I think should be set to `$_config.queryConfig` so that it's enabled in the ruler too.

PR #261, opened 2021-02-25
Add alert for etcd memory limits close

gouthamve:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master / unreleased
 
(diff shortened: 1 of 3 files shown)


pracucci (approved):
LGTM, but should that belong to the cortex-mixin? This is quite a generic alert, not specific to Cortex usage 🤔

gouthamve:
> This is quite a generic alert, not specific to Cortex usage

True, but this doesn't belong in the official etcd-mixin as they don't expect etcd to run in k8s. This is the best place I can think of :)

PR #273, opened 2021-03-16
Update Cortex scaling dashboard:

tomwilkie:
- Aim for 1.5M series per ingester
- Use recording rules to make some queries faster.

### DIFF
--- a/cortex-mixin/dashboards/scaling.libsonnet
+++ b/cortex-mixin/dashboards/scaling.libsonnet
@@ -34,11 +34,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
               or
             label_replace(
               sum by (cluster, namespace) (
-                4 * cortex_ingester_memory_series{cluster=~"$cluster", namespace=~"$namespace", job=~".+/ingester"}
-                  *
-                cortex_ingester_chunk_size_bytes_sum{cluster=~"$cluster", namespace=~"$namespace", job=~".+/ingester"}
-                  /
-                cortex_ingester_chunk_size_bytes_count{cluster=~"$cluster", namespace=~"$namespace", job=~".+/ingester"}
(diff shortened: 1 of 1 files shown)


cortex-mixin/dashboards/scaling.libsonnet
```diff
@@ -21,24 +21,20 @@ local utils = import 'mixin-utils/utils.libsonnet';
...
             )
               or
             label_replace(
               sum by (cluster, namespace) (
-                4 * cortex_ingester_memory_series{cluster=~"$cluster", namespace=~"$namespace", job=~".+/ingester"}
-                  *
-                cortex_ingester_chunk_size_bytes_sum{cluster=~"$cluster", namespace=~"$namespace", job=~".+/ingester"}
-                  /
-                cortex_ingester_chunk_size_bytes_count{cluster=~"$cluster", namespace=~"$namespace", job=~".+/ingester"}
+                increase(cortex_ingester_tsdb_storage_blocks_bytes{cluster=~"$cluster", namespace=~"$namespace", job=~".+/ingester"}[24h])
```
pracucci, line 37: I don't think this is correct. `cortex_ingester_tsdb_storage_blocks_bytes` is the size of compacted blocks on disk (but not deleted by TSDB retention yet) while I guess you want in-memory series. We don't have a specific metric for that. Should we use the go heap or container working set memory?

tomwilkie: How long do we keep blocks on disk in the ingesters - 96h?  I suggest we try and size memcached proportional to size of the blocks on disk - say, 1/4 of that.

tomwilkie:
Pretty sure this was all covered in https://github.com/grafana/cortex-jsonnet/pull/278

PR #279, opened 2021-03-23
Increased CortexRequestErrors alert severity

pracucci:
**What this PR does**:
The `CortexRequestErrors` alert looks to be quite critical. If it triggers, it means > 1% of requests to a specific route are failing for 15 consecutive minutes.

What's the sentiment if we raise it to `critical`?

N/A

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
   git fetch origin
   git branch -u origin/main main
   ```
-* [CHANGE] `CortexRequestErrors` alert severity raised from `warning` to `critical`. #278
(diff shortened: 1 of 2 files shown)


jtlisi (approved):
I'm 100% on board with this change. Before the the SLO alerts were added this alert was critical. One issue with the SLO alerts is it's difficult to tel if it was triggered by latency or errors. That means you can't really silence it for an intractable latency issue without being concerned an error issue could trigger later. This alert helps fill the gap.

pstibrany (approved):

gouthamve (approved):

PR #284, opened 2021-03-30
Pre-compute aggregations to optimize scaling recording rules

pracucci:
**What this PR does**:
Based on feedback received in https://github.com/grafana/cortex-jsonnet/pull/283#issuecomment-809469271, in this PR I'm proposing to pre-compute aggregations to try to optimize the scaling recording rules.

I haven't splitted recording rules into multiple groups in this PR in order to keep the diff easier to review.

N/A

**Checklist**

### DIFF
--- a/cortex-mixin/recording_rules.libsonnet
+++ b/cortex-mixin/recording_rules.libsonnet
@@ -198,6 +198,30 @@ local utils = import 'mixin-utils/utils.libsonnet';
             |||,
           },
           {
+            // Convenience rule to get the CPU utilization for both a deployment and a statefulset.
+            record: 'cluster_namespace_deployment:container_cpu_usage_seconds_total:sum_rate',
+            expr: |||
+              sum by (cluster, namespace, deployment) (
+                label_replace(
+                  node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate,
+                  "deployment", "$1", "pod", "(.*)-(?:([0-9]+)|([a-z0-9]+)-([a-z0-9]+))"
+                )
+              )
+            |||,
+          },
+          {
+            // Convenience rule to get the CPU request for both a deployment and a statefulset.
+            record: 'cluster_namespace_deployment:kube_pod_container_resource_requests_cpu_cores:sum',
+            expr: |||
+              sum by (cluster, namespace, deployment) (
+                label_replace(
+                  kube_pod_container_resource_requests_cpu_cores,
+                  "deployment", "$1", "pod", "(.*)-(?:([0-9]+)|([a-z0-9]+)-([a-z0-9]+))"
+                )
+              )
+            |||,
+          },
+          {
             // Jobs should be sized to their CPU usage.
             // We do this by comparing 99th percentile usage over the last 24hrs to
             // their current provisioned #replicas and resource requests.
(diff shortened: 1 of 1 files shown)


cortex-mixin/recording_rules.libsonnet
```diff
@@ -209,20 +233,32 @@ local utils = import 'mixin-utils/utils.libsonnet';
...
                   *
-                quantile_over_time(0.99,
-                  sum by (cluster, namespace, deployment) (
-                    label_replace(
-                      node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate,
-                      "deployment", "$1", "pod", "(.*)-(?:([0-9]+)|([a-z0-9]+)-([a-z0-9]+))"
-                    )
-                  )[24h:5m]
-                )
+                quantile_over_time(0.99, cluster_namespace_deployment:container_cpu_usage_seconds_total:sum_rate[24h:5m])
```
tomwilkie, line 236: Not shure the `[24h:5m]` here is the right syntax - you don't need a subquery, a plain range vector selector will do (ie just `[24h]`)

cortex-mixin/recording_rules.libsonnet
```diff
@@ -239,21 +275,9 @@ local utils = import 'mixin-utils/utils.libsonnet';
...
                   *
-                quantile_over_time(0.99,
-                  sum by (cluster, namespace, deployment) (
-                    label_replace(
-                      container_memory_usage_bytes,
-                      "deployment", "$1", "pod", "(.*)-(?:([0-9]+)|([a-z0-9]+)-([a-z0-9]+))"
-                    )
-                  )[24h:5m]
-                )
+                quantile_over_time(0.99, cluster_namespace_deployment:container_memory_usage_bytes:sum[24h:5m])
```
tomwilkie, line 278: Ditto here.

cortex-mixin/recording_rules.libsonnet
beorn7: The syntax still works. `[24h:5m]` will evaluate `cluster_namespace_deployment:container_cpu_usage_seconds_total:sum_rate` every 5m over the last 24h, thus 288 times.

`[24h]` will not evaluate anything but retrieve all the samples `cluster_namespace_deployment:container_cpu_usage_seconds_total:sum_rate` has for the last 24h. If the rule evaluation interval is 1m, that is 1440 samples.

I'm not sure what's faster, but since the latter just retrieves the time series data directly (which has to be processed anyway for the evaluation of the above), I could imagine the latter is faster, despite more data points being considered for the `quantile_over_time`.

The more important question is what the query should accomplish. I imagine the original query picked a resolution of 5m because it's barely enough to meaningfully calculate that 99th percentile, but ideally it would have had a higher resolution, but that was costly because of the subquery.

So I guess the `[24h]` form might not just be faster but also better for what we want to accomplish here.

beorn7 (changes requested):
I'd try the `[24h]` first. If that works out, great. Otherwise, we can play with the `[24h:5m]`.

pracucci:
Thanks @beorn7 for your detailed reply. I see your point. I removed the `:5m` step.

beorn7 (approved):

beorn7:
Let's see how it works out.

PR #285, opened 2021-03-31
Do not use deprecated Alertmanager cluster flags

pracucci:
**What this PR does**:
In this PR I'm replacing the deprecated Alertmanager cluster flags with new ones.

N/A

**Checklist**

### DIFF
--- a/cortex/alertmanager.libsonnet
+++ b/cortex/alertmanager.libsonnet
@@ -61,8 +61,8 @@
       container.withArgsMixin(
         $.util.mapToFlags($.alertmanager_args) +
         if isHA then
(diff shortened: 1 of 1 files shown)


gotjosh (approved):
LGTM - Could you please add a changelog entry?

pracucci:
> Could you please add a changelog entry?

Discussed offline. The change in this PR is expected to have no impact on the Cortex operator (we're switching from one flag to another flag which configures the exact same) so doesn't look we need a CHANGELOG entry for this.

gotjosh:
Disregard my changelog entry comment - it looks like we only include visible changes.

PR #286, opened 2021-03-31
fix: Update ksonnet-util vendor lock

jdbaldry:
**What this PR does**:

The previous version `c19a92e586a6752f11745b47f309b13f02ef7147` is
incompatible with the library in its current form. For example in
`tsdb.libsonnet` L81, we use `pvc.new('ingester-pvc')` but at the
locked version, in `ksonnet-util/kausal.libsonnet` the `pvc.new`
function takes no arguments.

**Checklist**

### DIFF
--- a/cortex/jsonnetfile.lock.json
+++ b/cortex/jsonnetfile.lock.json
@@ -38,8 +38,8 @@
           "subdir": "ksonnet-util"
         }
       },
-      "version": "c19a92e586a6752f11745b47f309b13f02ef7147",
-      "sum": "LKsTTBcH8TXX5ANgRUu5I7Y1tf5le4nANFV3/W53I+c="
+      "version": "8fa7669cc7b1b1822eb0220f2eda9c6aaa5c5119",
+      "sum": "/l/RofjusGrnNpJMD0ST+jDgtARyjvBP5vC7kEjPoQI="
     },
     {
       "source": {


pracucci (approved):
Ok, but shouldn't have break our jsonnet compilation?

jdbaldry:
I believe this won't have broken in our internal cluster because we actually vendor a different version (https://github.com/grafana/jsonnet-libs/tree/4d4b5b1ce01003547a110f93cc86b8b7afb282a6/ksonnet-util).

I completely understand the need to understand this better. I think we would benefit from having something like a `tk eval` recipe in a Makefile that evaluates the Jsonnet in the repo with specific configurations to validate that everything works.

I came across the incompatibility following the instructions in the `README.md` trying to manually test a change I'd like to make with regards to storage classes. To reproduce:

```console
$ git checkout main
$ cd $(git rev-parse --show-toplevel)/cortex
$ jb install # install cortex dependencies
$ jb install github.com/jsonnet-libs/k8s-alpha/1.18 # install k8s-alpha
$ mkdir -p lib
$ cat <<EOF > lib/k.libsonnet
(import 'github.com/jsonnet-libs/k8s-alpha/1.18/main.libsonnet')
+ (import 'github.com/jsonnet-libs/k8s-alpha/1.18/extensions/kausal-shim.libsonnet')
EOF
$ cat <<EOF > test.jsonnet
(import 'cortex.libsonnet') {
  _config+:: {
    bigtable_instance:: null,
    bigtable_project:: null,
    blocks_storage_bucket_name: 'test',
    namespace: 'test',
    storage_engine: 'blocks',
    storage_backend: 'gcp',
  },
}
EOF
$ jsonnet -J lib -J vendor test.jsonnet
```

pracucci (approved):
Good job spotting it! 👏  I agree having a Makefile recipe to run in CI would be great but let's not block on that.

jtlisi (approved):
LGTM, can second I ran into this issue recently setting up a local cluster

jdbaldry:
Thanks both for taking a look. Unfortunately, I'm not able to merge. Would I be able to be authorized?

pracucci:
> Thanks both for taking a look. Unfortunately, I'm not able to merge. Would I be able to be authorized?

Sure, doing it!

PR #287, opened 2021-04-02
Add function to customize compactor statefulset

pracucci:
**What this PR does**:
We have an use case where we need a custom compactor deployment. Similarly to what we did with `newIngesterStatefulSet()` function, in this PR I'm proposing to add `newCompactorStatefulSet()` function to allow to create a 2nd statefulset for the compactor.

I've also updated the compactor dashboards to allow to customize the compactor job name, like we already do for other services (eg. see the ingester one).

N/A

**Checklist**

### DIFF
--- a/cortex-mixin/dashboards/compactor-resources.libsonnet
+++ b/cortex-mixin/dashboards/compactor-resources.libsonnet
@@ -13,7 +13,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
         $.containerMemoryWorkingSetPanel('Memory (workingset)', 'compactor'),
       )
       .addPanel(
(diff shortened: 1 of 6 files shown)


cortex-mixin/dashboards/compactor-resources.libsonnet
```diff
@@ -13,7 +13,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
         $.containerMemoryWorkingSetPanel('Memory (workingset)', 'compactor'),
```
pracucci, line 13: No change here because this function wants the container name not the job name (same for CPU panel). The container name doesn't change when you customize the statefulset.

annanay25 (approved):
LGTM!

PR #288, opened 2021-04-02
Fixed CortexCompactorRunFailed threshold

pracucci:
**What this PR does**:
Due to interpolation, the computed `increase(cortex_compactor_runs_failed_total[2h])` may be slight higher than 1 (eg. `1.001`) so checking for `>= 2` (instead of `> 1`) fixes such false positives.

N/A

**Checklist**


sandeepsukhani (approved):
LGTM

PR #289, opened 2021-04-02
Added Cortex Rollout progress dashboard

pracucci:
**What this PR does**:
While rolling out new Cortex versions I typically keep an eye on it switching between multiple dashboards. In this PR I'm proposing a new dashboard "Rollout progress" aiming to show key metrics while rolling out Cortex:

Screenshot 2021-04-02 at 20 25 56

The dashboard uses the "new" Grafana grid system to place panels so I can't use the Grafana mixin vendored by `cortex-mixin`. Because of this I had to write a bit more template than usual. Future work may be done to clean it up, but I preferred to share the dashboard early since I think may be useful.

N/A

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
+* [FEATURE] Added "Cortex / Rollout progress" dashboard. #289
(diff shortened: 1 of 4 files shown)


pstibrany (approved):
Looks great. Let's get it in to see it working, and finetune from there if needed.

PR #291, opened 2021-04-08
Add querier_service_ignored_labels

tsanghi:
**What this PR does**:

`querier_service_ignored_labels` is used in `gossip.libsonnet` but not defined in `querier.libsonnet`.


pstibrany (approved):
Well spotted, thank you!

PR #292, opened 2021-04-20
Simplify compactor alerts

gouthamve:
We should simply alert on things not having run since X.

### DIFF
--- a/cortex-mixin/alerts/compactor.libsonnet
+++ b/cortex-mixin/alerts/compactor.libsonnet
@@ -6,11 +6,9 @@
         {
           // Alert if the compactor has not successfully cleaned up blocks in the last 24h.
           alert: 'CortexCompactorHasNotSuccessfullyCleanedUpBlocks',
-          'for': '15m',
+          'for': '1h',
           expr: |||
             (time() - cortex_compactor_block_cleanup_last_successful_run_timestamp_seconds > 60 * 60 * 24)
-            and
-            (cortex_compactor_block_cleanup_last_successful_run_timestamp_seconds > 0)
           |||,
           labels: {
             severity: 'critical',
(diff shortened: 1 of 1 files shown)


cortex-mixin/alerts/compactor.libsonnet
```diff
@@ -20,17 +18,17 @@
           },
         },
         {
-          // Alert if the compactor has not successfully cleaned up blocks since its start.
-          alert: 'CortexCompactorHasNotSuccessfullyCleanedUpBlocksSinceStart',
-          'for': '24h',
+          // Alert if the compactor has not successfully run compaction in the last 6h.
+          alert: 'CortexCompactorHasNotSuccessfullyRunCompaction',
```
pracucci, line 22: The name  says "run compaction" but it checks blocks cleanup. Not sure what you wanna check here.

cortex-mixin/alerts/compactor.libsonnet
```diff
@@ -20,17 +18,17 @@
...
+          'for': '1h',
           expr: |||
-            cortex_compactor_block_cleanup_last_successful_run_timestamp_seconds == 0
+            time() - cortex_compactor_block_cleanup_last_successful_run_timestamp_seconds > 60 * 60 * 6
```
pracucci, line 25: As it is right now, it's the same as `CortexCompactorHasNotSuccessfullyCleanedUpBlocks` just with a different threshold, so this one will trigger first and the other alert is useless. I think keeping this alert with a 6h threshold and a for of 1h is fine (I checked in our infrastructure over the last 3 weeks and would have never triggered).

cortex-mixin/alerts/compactor.libsonnet
```diff
@@ -6,11 +6,9 @@
         {
           // Alert if the compactor has not successfully cleaned up blocks in the last 24h.
           alert: 'CortexCompactorHasNotSuccessfullyCleanedUpBlocks',
-          'for': '15m',
+          'for': '1h',
           expr: |||
             (time() - cortex_compactor_block_cleanup_last_successful_run_timestamp_seconds > 60 * 60 * 24)
```
pracucci, line 11: I think we can lower the threshold to 6 hours. I've checked our infra and a 6h threshold looks large and good enough.

```suggestion
            (time() - cortex_compactor_block_cleanup_last_successful_run_timestamp_seconds > 60 * 60 * 6)
```

**If you pick up this suggestion, we need to update the `message` below too.**

cortex-mixin/alerts/compactor.libsonnet
```diff
@@ -6,11 +6,9 @@
         {
           // Alert if the compactor has not successfully cleaned up blocks in the last 24h.
```
pracucci, line 7:
```suggestion
          // Alert if the compactor has not successfully cleaned up blocks in the last 6h.
```

pracucci (approved):
LGTM (modulo a comment about threshold, which I believe we can improve).

PR #293, opened 2021-04-20
Use new compaction interval metric in compaction failed alert.

cstyan:

**What this PR does**:
Changes compaction failed alert to use new `cortex_compactor_compaction_interval_seconds` metric, so that we can more accurately alert when two compactions in a row have failed, regardless of whether the compaction interval is 2h (the value currently used within the increase function of the alert) or not.

Not sure if there should be a changelog entry here.

**Checklist**

### DIFF
--- a/cortex-mixin/alerts/compactor.libsonnet
+++ b/cortex-mixin/alerts/compactor.libsonnet
@@ -65,7 +65,8 @@
           // Alert if compactor fails.
           alert: 'CortexCompactorRunFailed',
           expr: |||
-            increase(cortex_compactor_runs_failed_total[2h]) >= 2
+            (time() - cortex_compactor_last_successful_run_timestamp_seconds) > 
+            (cortex_compactor_compaction_interval_seconds * 2)
           |||,
           labels: {
             severity: 'critical',


pstibrany (approved):
LGTM, thanks!

pracucci (commented):
I think this PR conflicts with #294. We should have already covered the case in #294 and actually ,due to how compactor works and `cortex_compactor_last_successful_run_timestamp_seconds ` is updated, it may take longer than 2x compaction interval to complete a compaction run (and the compactor may be perfectly fine anyway).

pstibrany:
Can we go back to using `increase(cortex_compactor_runs_failed_total[2h])` but with interval defined in the metric? I don't think that's possible though :(

cstyan:
> Can we go back to using `increase(cortex_compactor_runs_failed_total[2h])` but with interval defined in the metric? I don't think that's possible though :(

Yeah, my understanding is that we can't do this in PromQL right now. @beorn7 would be able to confirm.

beorn7:
Are you talking about using the value of a metric rather than the duration literal `2h` in `[2h]`?

If that's the case, then the answer is: no, that's currently not possible in PromQL. I do think we should make that possible, but it's one of those "not as easy as it looks" problems. Right now, the PromQL engine can find out what time range the query will access in the TSDB, just from static analysis of the query. With allowing the value coming from another expression, you now have to evaluate that "inner query" first to find out what time range the "outer query" will need to access.

For reference: My brainstorming doc about timestamps and durations: https://docs.google.com/document/d/1jMeDsLvDfO92Qnry_JLAXalvMRzMSB1sBr9V7LolpYM/edit#heading=h.vmb7pe7hp12

pracucci:
May you confirm we can't query the last time (seconds) a metric has
increased, right?

beorn7:
I believe there is no straightforward way to do this within PromQL. But perhaps an expert can prove me wrong here?

There is the `timestamp()` function to access the actual timestamp, but it only works on an instant vector. So whenever you apply it to an expression (e.g. to filter for the last sample before an increase or the first after an increase), it will only yield the evaluation timestamp, not the timestamp of any samples that fed into the expression.

tleopold:

PR #294, opened 2021-04-21
Fix CortexCompactorHasNotSuccessfullyRunCompaction to avoid false positives

pracucci:
**What this PR does**:
Yesterday we merged #292 but I just realised it suffers false positives. The reason is that a compaction run is marked as done once there's no more work to do for the compactor. Because of this it could take even a long time (several hours) while the compactor is working flawlessly  as expected. Moreover, after startup the metric `cortex_compactor_last_successful_run_timestamp_seconds` value will be `0` until the first compaction run as completed (which, again, could take hours).

In this PR:
1. Increase threshold to 24h
2. Cover the case of the metric == 0 right after startup (while the 1st compaction run is on-going)

N/A

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * [FEATURE] Added "Cortex / Rollout progress" dashboard. #289 #290
 * [ENHANCEMENT] Added `newCompactorStatefulSet()` function to create a custom statefulset for the compactor. #287
 * [ENHANCEMENT] Added option to configure compactor job name used in dashboards and alerts. #287
+* [ENHANCEMENT] Added `CortexCompactorHasNotSuccessfullyRunCompaction` alert. #292 #294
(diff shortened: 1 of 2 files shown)


pstibrany (approved):

gouthamve:
How long of a compaction run is considered bad? Is a compaction run taking 24hrs bad? Is a run taking 6hrs bad?

pracucci:
> How long of a compaction run is considered bad? Is a compaction run taking 24hrs bad? Is a run taking 6hrs bad?

There's no easy anwer. A compactor run loops until there's work to do. It may take forever and still catching up with the work, but run in a forever loop because as soon as it completes to compact a block there's more work to do. Realistically, we've observed 24h should be a threshold long enough. If compactor doesn't fully catch up in 24h, then you need to scale it up (vertically or horizontally).

gouthamve (approved):
LGTM

PR #295, opened 2021-04-22
Add `query-scheduler.libsonnet`

pstibrany:
**What this PR does**: This PR adds `query-scheduler.libsonnet` for configuring query-scheduler.

**Checklist**

### DIFF
--- a/cortex/images.libsonnet
+++ b/cortex/images.libsonnet
@@ -17,6 +17,7 @@
     flusher: self.cortex,
     ruler: self.cortex,
     store_gateway: self.cortex,
+    query_scheduler: $._images.cortex,
 
     cortex_tools: 'grafana/cortex-tools:v0.4.0',
     query_tee: 'quay.io/cortexproject/query-tee:v1.8.0',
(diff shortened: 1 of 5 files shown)


pracucci (commented):
Instead of having to include the source file to enable it, can we just have a `cortex_query_scheduler_enabled` in the config? I think it would be more practical to enable/disable.

pstibrany:
> Instead of having to include the source file to enable it, can we just have a `cortex_query_scheduler_enabled` in the config? I think it would be more practical to enable/disable.

You're right. Please take a look again.

cortex/images.libsonnet
```diff
@@ -17,6 +17,7 @@
     flusher: self.cortex,
     ruler: self.cortex,
     store_gateway: self.cortex,
+    query_scheduler: $._images.cortex,
```
pracucci, line 20:
```suggestion
    query_scheduler: self.cortex,
```

pracucci (approved):
Thanks for working on it! LGTM (modulo a nit) 🙏

cortex/images.libsonnet
pstibrany: 🤦 Thanks!

pstibrany: (Although it worked the other way too ;-))

PR #301, opened 2021-05-11
feat: Allow configuration of ring members in gossip alerts

jdbaldry:
**What this PR does**:
Allows library consumers to configure the names of jobs that are members of the gossip ring which is used by the CortexGossipMembersMismatch alert.

**Checklist**

### DIFF
--- a/cortex-mixin/config.libsonnet
+++ b/cortex-mixin/config.libsonnet
@@ -33,7 +33,9 @@
       query_frontend: '(query-frontend.*|cortex$)',  // Match also custom query-frontend deployments.
       query_scheduler: 'query-scheduler.*',  // Not part of single-binary. Match also custom query-scheduler deployments.
       table_manager: '(table-manager|cortex$)',
-      ring_members: '(compactor|distributor|ingester|querier|ruler|store-gateway|cortex)',
+      // ingester-.* accommodates multiple ingester StatefulSets or Deployments.
+      // cortex$ prevents matching the cortex-gateway.
+      ring_members: '(compactor|distributor|ingester-.*|querier|ruler|store-gateway|cortex$)',
       store_gateway: '(store-gateway|cortex$)',
       gateway: '(gateway|cortex-gw|cortex-gw-internal)',
       compactor: 'compactor.*',  // Match also custom compactor deployments.
--- a/cortex-mixin/config.libsonnet
+++ b/cortex-mixin/config.libsonnet
@@ -33,9 +33,13 @@
       query_frontend: '(query-frontend.*|cortex$)',  // Match also custom query-frontend deployments.
       query_scheduler: 'query-scheduler.*',  // Not part of single-binary. Match also custom query-scheduler deployments.
       table_manager: '(table-manager|cortex$)',
+      // If non-null, this list is joined together to form the regexp for the ring_members matcher value.
+      // to completely, override the matcher value, you can make this null and override the `ring_members`
+      // field instead of this.
       // ingester-.* accommodates multiple ingester StatefulSets or Deployments.
       // cortex$ prevents matching the cortex-gateway.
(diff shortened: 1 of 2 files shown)


simonswine:
I have tried to fix the same problem: https://github.com/grafana/cortex-jsonnet/pull/305/files

simonswine:
I think it is probably preferable to make the components configurable

cortex-mixin/config.libsonnet
```diff
@@ -33,6 +33,7 @@
       query_frontend: '(query-frontend|cortex$)',
       query_scheduler: 'query-scheduler',  // Not part of single-binary.
       table_manager: '(table-manager|cortex$)',
+      ring_members: '(compactor|distributor|ingester|querier|ruler|store-gateway|cortex)',
```
simonswine, line 36: I would prefer ring members being an array instead.

Also be careful with the `ingester` vs `ingester.*`

simonswine (changes requested):

cortex-mixin/config.libsonnet
jdbaldry: Good catch on the `ingester` -> `ingester.*` change. I didn't mean to do that but thinking about it, I'm also not sure why it would be `ingester.*` when we relabel jobs based upon the Kubernetes name label. I probably need to change `cortex` to be `cortex$` to avoid potential naming clashes too.

simonswine: Guess: It might have to with some migrations. I think there were some cases were ingesters were migrated from one stateful set to another via `ingesters-temp` one.

jdbaldry: I've update the list to correctly match all ingester workloads and avoid matching the cortex-gateway. I've kept this as a string for now as all the other job configuration values are also strings. Is there a specific reason to prefer an array for ring_members?

cortex-mixin/config.libsonnet
```diff
@@ -33,6 +33,13 @@
       query_frontend: '(query-frontend|cortex$)',
       query_scheduler: 'query-scheduler',  // Not part of single-binary.
       table_manager: '(table-manager|cortex$)',
+      // If non-null, this list is joined together to form the regexp for the ring_members matcher value.
+      // to completely, override the matcher value, you can make this null and override the `ring_members`
+      // field instead of this.
+      // ingester-.* accommodates multiple ingester StatefulSets or Deployments.
+      // cortex$ prevents matching the cortex-gateway.
+      ring_members_list: ['compactor', 'distributor', 'ingester-.*', 'querier', 'ruler', 'store-gateway', 'cortex$'],
```
pracucci, line 41: PromQL regexs are anchored, so the `cortex$` can be changed into `cortex`

cortex-mixin/config.libsonnet
```diff
@@ -33,6 +33,13 @@
...
+      ring_members: if self.ring_members_list != null then '(%s)' % std.join('|', self.ring_members_list) else '',
```
pracucci, line 42: I'm wondering if we could rename `ring_members_list` into `ring_members` and get rid of this `ring_members_list`. 

Keeping the 2 variables could end up to some misunderstanding: if you configure both, `ring_members_list` will be ignored. The logic in this `ring_members` could be moved to a function instead. WDYT?

cortex-mixin/config.libsonnet
```diff
@@ -33,6 +33,7 @@
       query_frontend: '(query-frontend|cortex$)',
       query_scheduler: 'query-scheduler',  // Not part of single-binary.
       table_manager: '(table-manager|cortex$)',
+      ring_members: ['compactor', 'distributor', 'ingester-.*', 'querier', 'ruler', 'store-gateway', 'cortex'],
```
jdbaldry, line 36: Note, this presently removes the admin-api which is GEM specific.

simonswine (approved):
LGTM

PR #302, opened 2021-05-11
Default dashboards to Cortex blocks storage only

pracucci:
**What this PR does**:
With the default config, dashboards both display panels related to Cortex chunks and blocks storage. Given most Cortex users have moved to blocks I propose to change the default to blocks only.

N/A

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## master / unreleased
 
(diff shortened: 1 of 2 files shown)


pstibrany (approved):
I don't mind the change, but how do we know that "most users have moved to blocks"? Based on issues and Slack, there is still plenty of chunks installations, even new ones since our Helm chart uses that.

pracucci:
> I don't mind the change, but how do we know that "most users have moved to blocks"? Based on issues and Slack, there is still plenty of chunks installations, even new ones since our Helm chart uses that.

You're right, we don't have numbers. Let's say then we do this change to push the blocks storage forward :=

PR #304, opened 2021-05-12
Do not use deprecated flag -limits.per-user-override-config

pracucci:
**What this PR does**:
The flag `-limits.per-user-override-config` will be removed in Cortex 1.9 (RC is already out), so in this PR I'm replacing it with the new `-runtime-config.file`.

N/A

**Checklist**


pstibrany (approved):
Thanks!

PR #308, opened 2021-05-17
Changed alertmanager config to use the new storage config

pracucci:
**What this PR does**:
Similarly to https://github.com/grafana/cortex-jsonnet/pull/306, in this PR I'm switching the ruler to use the new storage config. I'm also taking the opportunity to do a couple of other changes:

1. Remove configdb support from ruler and alertmanager: wasn't working even before, so I guess none was using it
2. Do not auto-configure `alertmanager_s3_bucket_name` and `ruler_s3_bucket_name` (like we already do for GCS)

N/A

**Checklist**


gotjosh (approved):

pstibrany (approved):
Looks good.

PR #317, opened 2021-05-31
cortex-mixin: Make `cluster_namespace_deployment:kube_pod_container_resource_requests_{cpu_cores,memory_bytes}:sum` backwards compatible with `kube-state-metrics` v2.0.0

jvrplmlmn:
**What this PR does**:

This updates the recording rules to make them compatible with kube-state-metrics v2.0.0
which introduces some breaking changes in some metric names.

With kube-state-metrics v2.0.0:
- `kube_pod_container_resource_requests_cpu_cores` becomes `kube_pod_container_resource_requests{resource="cpu"}`
- `kube_pod_container_resource_requests_memory_bytes` becomes `kube_pod_container_resource_requests{resource="memory"}`

Edit: I have taken a look at `kube-state-metrics` [CHANGELOG](https://github.com/kubernetes/kube-state-metrics/blob/master/CHANGELOG.md) and since [v1.4.0](https://github.com/kubernetes/kube-state-metrics/blob/master/CHANGELOG.md#v140-rc0--2018-08-06) the _new_ metrics that I'm referencing here are available. Internally I have double checked that the new expression is compatible, as it returns the same data as the old one.

N/A

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * [CHANGE] `namespace` template variable in dashboards now only selects namespaces for selected clusters. #311
 * [CHANGE] Alertmanager: mounted overrides configmap to alertmanager too. #315
 * [CHANGE] Memcached: upgraded memcached from `1.5.17` to `1.6.9`. #316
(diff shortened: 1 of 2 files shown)


pracucci (commented):
I would add backward compatibility using `or` so we don't introduce a breaking change.

pracucci (approved):
LGTM!

PR #319, opened 2021-06-02
refactor: config for job aggregation strings

darrenjaneczek:
**What this PR does**:
(Updated based on suggestions below)
- to make it easier to override, define "cluster_namespace_job" and "cluster, namespace, job" in terms of a new list in $._config: `job_labels: ['cluster', 'namespace', 'job']`
  - This replaces many hard-coded uses of these labels
- for consistency, based on feedback, we now follow a similar pattern for clusters:  `cluster_labels: ['cluster', 'namespace']`
- the resulting string values are defined using a new groups.libsonnet file as `$._group_config`, which gets mixed in where it is needed.
- a hard-coded group-by used `namespace` hard-coded in the query -- it has been updated to use the composition from `$._cluster_labels`. 
- deprecated `_config.alert_aggregation_labels`, since it is now a function of `$._cluster_labels`
  - there is a warning message if this old value is still being directly overridden; it still works (for now)
  - encouraged approach is to override the `cluster_labels` list instead, which is what will happen if `alert_aggregation_labels` is not overridden. 

The resulting output does not change (unless any of the new lists in config are overridden), except in exactly one case where it was recommended to replace a hard-coded reference from `namespace` to `cluster, namespace`, composed from `$._config.cluster_labels`.

**Checklist**

### DIFF
--- a/cortex-mixin/config.libsonnet
+++ b/cortex-mixin/config.libsonnet
@@ -38,6 +38,11 @@
       compactor: 'compactor.*',  // Match also custom compactor deployments.
     },
 
+    // Aggregation strings related to "jobs"
+    job_aggregation_prefix: 'cluster_namespace_job',
+    job_aggregation_labels_recording_rules: 'cluster, namespace, job',
+    job_aggregation_labels_active_series: 'namespace',
+
     // Labels used to in alert aggregations - should uniquely identify
     // a single Cortex cluster.
     alert_aggregation_labels: 'cluster, namespace',
--- a/cortex-mixin/config.libsonnet
+++ b/cortex-mixin/config.libsonnet
@@ -38,14 +38,10 @@
       compactor: 'compactor.*',  // Match also custom compactor deployments.
     },
 
-    // Aggregation strings related to "jobs"
-    job_aggregation_prefix: 'cluster_namespace_job',
-    job_aggregation_labels_recording_rules: 'cluster, namespace, job',
-    job_aggregation_labels_active_series: 'namespace',
+    // Grouping labels, to uniquely identify and group by {jobs, clusters}
+    job_labels: ['cluster', 'namespace', 'job'],
+    cluster_labels: ['cluster', 'namespace'],
 
-    // Labels used to in alert aggregations - should uniquely identify
-    // a single Cortex cluster.
-    alert_aggregation_labels: 'cluster, namespace',
     cortex_p99_latency_threshold_seconds: 2.5,
 
     // Whether resources dashboards are enabled (based on cAdvisor metrics).
--- a/cortex-mixin/dashboards/writes.libsonnet
+++ b/cortex-mixin/dashboards/writes.libsonnet
@@ -13,10 +13,10 @@ local utils = import 'mixin-utils/utils.libsonnet';
       .addPanel(
         $.panel('Samples / s') +
         $.statPanel(
-            'sum(%(jobAggregationPrefix)s:cortex_distributor_received_samples:rate5m{%(job)s})' % {
+          'sum(%(jobAggregationPrefix)s:cortex_distributor_received_samples:rate5m{%(job)s})' % {
             job: $.jobMatcher($._config.job_names.distributor),
-            jobAggregationPrefix: $._config.job_aggregation_prefix
-          }, 
+            jobAggregationPrefix: $._config.job_aggregation_prefix,
+          },
           format='reqps'
         )
       )
@@ -29,7 +29,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
         ||| % {
           ingester: $.jobMatcher($._config.job_names.ingester),
           distributor: $.jobMatcher($._config.job_names.distributor),
-          labels: $._config.job_aggregation_labels_active_series
+          labels: $._config.job_aggregation_labels_active_series,
         }, format='short')
       )
       .addPanel(
--- a/cortex-mixin/dashboards/writes.libsonnet
+++ b/cortex-mixin/dashboards/writes.libsonnet
@@ -13,9 +13,8 @@ local utils = import 'mixin-utils/utils.libsonnet';
       .addPanel(
         $.panel('Samples / s') +
         $.statPanel(
-          'sum(%(jobAggregationPrefix)s:cortex_distributor_received_samples:rate5m{%(job)s})' % {
+          'sum(%(job_aggregation_prefix)s:cortex_distributor_received_samples:rate5m{%(job)s})' % $._config + {
             job: $.jobMatcher($._config.job_names.distributor),
-            jobAggregationPrefix: $._config.job_aggregation_prefix,
           },
           format='reqps'
         )
--- a/cortex-mixin/dashboards/writes.libsonnet
+++ b/cortex-mixin/dashboards/writes.libsonnet
@@ -13,9 +13,11 @@ local utils = import 'mixin-utils/utils.libsonnet';
       .addPanel(
         $.panel('Samples / s') +
         $.statPanel(
-          'sum(%(job_aggregation_prefix)s:cortex_distributor_received_samples:rate5m{%(job)s})' % $._config + {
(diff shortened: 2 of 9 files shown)


darrenjaneczek:
I am hoping for some feedback on the variable names, to ensure that they are appropriate.

56quarters (approved):
LGTM, but I'd wait for someone from the Cortex team

cortex-mixin/dashboards/writes.libsonnet
```diff
@@ -12,17 +12,24 @@ local utils = import 'mixin-utils/utils.libsonnet';
        })
       .addPanel(
         $.panel('Samples / s') +
-        $.statPanel('sum(cluster_namespace_job:cortex_distributor_received_samples:rate5m{%s})' % $.jobMatcher($._config.job_names.distributor), format='reqps')
+        $.statPanel(
+          'sum(%(jobAggregationPrefix)s:cortex_distributor_received_samples:rate5m{%(job)s})' % {
+            job: $.jobMatcher($._config.job_names.distributor),
+            jobAggregationPrefix: $._config.job_aggregation_prefix,
+          },
```
pracucci, line 19: [nit] If I'm not missing anything, you may simplify it as:

```suggestion
          'sum(%(job_aggregation_prefix)s:cortex_distributor_received_samples:rate5m{%(job)s})' % $._config + {
            job: $.jobMatcher($._config.job_names.distributor),
          },
```

cortex-mixin/dashboards/writes.libsonnet
```diff
@@ -12,17 +12,24 @@ local utils = import 'mixin-utils/utils.libsonnet';
...
         $.statPanel(|||
           sum(cortex_ingester_memory_series{%(ingester)s}
-          / on(namespace) group_left
-          max by (namespace) (cortex_distributor_replication_factor{%(distributor)s}))
+          / on(%(labels)s) group_left
+          max by (%(labels)s) (cortex_distributor_replication_factor{%(distributor)s}))
         ||| % {
           ingester: $.jobMatcher($._config.job_names.ingester),
           distributor: $.jobMatcher($._config.job_names.distributor),
+          labels: $._config.job_aggregation_labels_active_series,
```
pracucci, line 32: We should generalise it and not make it "active series" specific. What we want here is to aggregate by Cortex cluster. Also the config variable is named "job_" but there's not job involved in the aggregation here.

An idea could be renaming the fields as:
- job_aggregation_labels_recording_rules > group_by_job_labels (because they're the labels to use to group by job)
- job_aggregation_labels_active_series > group_by_cluster_labels (because they're the labels to group by Cortex cluster). This should also be merged with `alert_aggregation_labels`, so changing this panel to actually group by `cluster, namespace` by default (it's correct for this use case)

cortex-mixin/config.libsonnet
```diff
@@ -38,6 +38,11 @@
       compactor: 'compactor.*',  // Match also custom compactor deployments.
     },
 
+    // Aggregation strings related to "jobs"
+    job_aggregation_prefix: 'cluster_namespace_job',
```
pracucci, line 42: This can (and probably should) be auto-generated based on the value of `job_aggregation_labels_recording_rules`. It's about doing a string substitution:

```suggestion
    job_aggregation_prefix: std.strReplace(self. job_aggregation_labels_recording_rules, ', ', '_'),
```

pracucci (commented):
Thanks for working on this! I left few comments 🙏

cortex-mixin/config.libsonnet
darrenjaneczek: I like this idea. Perhaps we can take it a step further, and define `_config.job_aggregation_labels` as a list, and that list is used to generate both the `","`-separated and `"_"`-separated strings; I assume those generated strings wouldn't need to appear in the config, only the list that they are generated from.

cortex-mixin/dashboards/writes.libsonnet
darrenjaneczek: If I remove/rename `alert_aggregation_labels`, is there anything we can do for consumers of this who might have overridden this variable? Or is this a low risk?

darrenjaneczek: After merging the variables, the output is slightly different (as expected). `cluster, namespace` replaces the sole `namespace`.

```diff
diff /home/darren/testOutMain/dashboards/cortex-writes.json /home/darren/testOutBranch/dashboards/cortex-writes.json
138c138
<                         "expr": "sum(cortex_ingester_memory_series{cluster=~\"$cluster\", job=~\"($namespace)/(ingester.*|cortex$)\"}\n/ on(namespace) group_left\nmax by (namespace) (cortex_distributor_replication_factor{cluster=~\"$cluster\", job=~\"($namespace)/(distributor|cortex$)\"}))\n",
---
>                         "expr": "sum(cortex_ingester_memory_series{cluster=~\"$cluster\", job=~\"($namespace)/(ingester.*|cortex$)\"}\n/ on(cluster, namespace) group_left\nmax by (cluster, namespace) (cortex_distributor_replication_factor{cluster=~\"$cluster\", job=~\"($namespace)/(distributor|cortex$)\"}))\n",
```

darrenjaneczek: Also, `alert_aggregation_labels` has been marked as deprecated. If someone is overriding that value, it will still work as they expect, but a std.trace deprecation warning will print out. In the future, we can consider replacing this warning with an error when all references using that variable are removed.

cortex-mixin/config.libsonnet
darrenjaneczek: I simplified the _config aspect to only define the two lists that these prefix/group-by strings are generated from.
This required adding a new groups.libsonnet file. This helps with generating the strings consistently, and making them available along with the base _config.

darrenjaneczek:
@pracucci , with the `[cluster, namespace]` list (`cluster_labels`) defining which labels identify a cluster, does it make sense to seek out references to `cluster_namespace*` and replace them with the corresponding prefix (`group_prefix_clusters`)?

cortex-mixin/groups.libsonnet
```diff
@@ -0,0 +1,45 @@
...
+    // Each group-by label list is `, `-separated and unique identifies
+    group_by_job: makeGroupBy($._config.job_labels),
+    group_by_cluster: group_by_cluster,
+  },
+
+  // The following works around the deprecation of `$._config.alert_aggregation_labels`
+  // - If an override of that value is detected, a warning will be printed
+  // - If no override was detected, it will be set to the `group_by_cluster` value,
+  //   which will replace it altogether in the future.
+  local alert_aggregation_labels_override = (
```
pracucci, line 21: Good job here! I'm just wondering if we actually need this complexity. What we typically do is to introduce a breaking change (eg. stop using `alert_aggregation_labels`) and document it in the CHANGELOG. This mixin policy doesn't guarantee backward compatibility.

pracucci (approved):
LGTM (modulo a comment about `alert_aggregation_labels_override`). Please also add a CHANGELOG entry to explain the new config options.

pracucci:
> @pracucci , with the `[cluster, namespace]` list (`cluster_labels`) defining which labels identify a cluster, does it make sense to seek out references to `cluster_namespace*` string patterns and replace them with the corresponding prefix (`group_prefix_clusters`)?

I would start as smallest as possible because the potential replacement is huge. If this PR solves your use case, I would start from here and work on it on incremental PRs.

cortex-mixin/groups.libsonnet
darrenjaneczek: I was concerned that my change could cause a phantom failure for someone, where their custom override would just stop working without any warning/error.

A plain old "`error` if `alert_aggregation_labels` is defined" would be less complex and solve my concern.

Let me know if/how you want this addressed in this PR.

pracucci: Let's not block on this. What you did looks correct, I'm just a bit worried about its maintenance, but not a blocker.

PR #329, opened 2021-06-16
Fixed --alertmanager.cluster.peers

pracucci:
**What this PR does**:
The `--alertmanager.cluster.peers` is expected to have multiple peers a single comma-separated string, instead of  specifying multiple times the CLI flag. This PR fixes it.

N/A

**Checklist**

### DIFF
--- a/cortex/alertmanager.libsonnet
+++ b/cortex/alertmanager.libsonnet
@@ -63,7 +63,7 @@
         $.util.mapToFlags($.alertmanager_args) +
         if isHA then
           ['--alertmanager.cluster.listen-address=[$(POD_IP)]:%s' % $._config.alertmanager.gossip_port] +
-          ['--alertmanager.cluster.peers=%s' % peer for peer in peers]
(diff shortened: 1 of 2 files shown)


cortex/alertmanager.libsonnet
```diff
@@ -63,7 +63,7 @@
         $.util.mapToFlags($.alertmanager_args) +
         if isHA then
           ['--alertmanager.cluster.listen-address=[$(POD_IP)]:%s' % $._config.alertmanager.gossip_port] +
-          ['--alertmanager.cluster.peers=%s' % peer for peer in peers]
+          ['--alertmanager.cluster.peers=%s' % std.join(',', peers)]
```
gotjosh, line 66: Is this a good time to also fix the double `--`?

gotjosh (approved):

cortex/alertmanager.libsonnet
pracucci: Not now, otherwise will trigger a rollout in all envs :)

PR #330, opened 2021-06-16
Add a simple playbook for ingester series limit alert.

cstyan:
Pretty minimal but better than nothing, what we talked about this morning @pracucci 


pracucci (approved):
👍  Thanks!

PR #340, opened 2021-06-25
Change min-step to 15s to show better detail.

### LINKED ISSUE #339: Query 'min step' of 1m loses fine detail
I find this setting problemmatic:
https://github.com/grafana/cortex-jsonnet/blob/b32d0421a0d5e6258dfb60e41a863c20e7e49170/cortex-mixin/dashboards/dashboard-utils.libsonnet#L87

It is shown in the Grafana UI as 'min step'.

[Here](https://grafana.com/docs/grafana/latest/datasources/prometheus/#using-__rate_interval-variable) it is documented that `$__rate_interval` will be floored at 4x this quantity, i.e. 4 minutes.  This makes it impossible to see fast transients in the data.

Further, Cortex query-frontend will quantise start and end times to a multiple of the step time, so dashboards only change every minute.

Since we are scraping at 15s I propose "min step" should be 15s.

bboreham:
**What this PR does**:
`$__rate_interval` will be floored at 4x this quantity, so 15s lets us see faster transients than the previous value of 1m.

Fixes #339 

**Checklist**
- [x] `CHANGELOG.md` updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * [CHANGE] Renamed `CortexCompactorRunFailed` to `CortexCompactorHasNotSuccessfullyRunCompaction`. #334
 * [CHANGE] Renamed `CortexInconsistentConfig` alert to `CortexInconsistentRuntimeConfig` and increased severity to `critical`. #335
 * [CHANGE] Increased `CortexBadRuntimeConfig` alert severity to `critical` and removed support for `cortex_overrides_last_reload_successful` metric (was removed in Cortex 1.3.0). #335
+* [CHANGE] Grafana 'min step' changed to 15s so dashboard show better detail. #340
 * [ENHANCEMENT] cortex-mixin: Make `cluster_namespace_deployment:kube_pod_container_resource_requests_{cpu_cores,memory_bytes}:sum` backwards compatible with `kube-state-metrics` v2.0.0. #317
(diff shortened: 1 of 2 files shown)


pracucci (approved):
Makes sense to me!

Unrelated from this PR: I've also noticed in the in the `queries.libsonnet` dashboard we've some queries with the rate time window hardcoded to `[1m]`. Should we change it to `[$__rate_interval]` like other dashboards?

bboreham:
> we've some queries with the rate time window hardcoded to `[1m]`

The main difference I'm aware of is that the window will get bigger as you zoom out with `[$__rate_interval]`.  So spikes will get more rounded off.  At small time ranges it will still hit the floor so no difference; at large time ranges `[1m]` is very spiky so a smoother picture is easier to read.

Empirically on my machine it jumps from `1m` to `2m` between 24h and 2 days, and both are pretty spiky, so I would agree we should change `queries.libsonnet` to `[$__rate_interval]` like other dashboards.

Incidentally I found [Beorn wrote an explanation](https://grafana.com/blog/2020/09/28/new-in-grafana-7.2-__rate_interval-for-prometheus-rate-queries-that-just-work/) which suggests that `interval` was intended to tell Grafana the scrape interval, so this PR is bringing the world into alignment.

RichiH (approved):

PR #341, opened 2021-07-01
Added playbook for CortexFrontendQueriesStuck and CortexSchedulerQueriesStuck

pracucci:
**What this PR does**:
Added playbook for CortexFrontendQueriesStuck and CortexSchedulerQueriesStuck. I tried to write down the most common root causes: anything else to add?

N/A

**Checklist**

### DIFF
--- a/cortex-mixin/docs/playbooks.md
+++ b/cortex-mixin/docs/playbooks.md
@@ -414,13 +414,13 @@ The procedure to investigate it is the same as the one for [`CortexSchedulerQuer
 
 ### CortexSchedulerQueriesStuck
 
-This alert fires if Cortex is queries are piling up in the query-scheduler.
+This alert fires if queries are piling up in the query-scheduler.
 
 How it **works**:
 - A query-frontend API endpoint is called to execute a query
 - The query-frontend enqueues the request to the query-scheduler
-- The query-scheduler is responsible to dispatch enqueued queries to idle querier workers
-- The querier runs the query, sends the response back directly to the query-frontend and notifies the query-scheduler
+- The query-scheduler is responsible for dispatching enqueued queries to idle querier workers
+- The querier runs the query, sends the response back directly to the query-frontend and notifies the query-scheduler that it can process another query
 
 How to **investigate**:
 - Are queriers in a crash loop (eg. OOMKilled)?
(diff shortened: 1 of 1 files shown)


aknuds1 (approved):
Looks good AFAICT

cortex-mixin/docs/playbooks.md
```diff
@@ -408,11 +408,31 @@ _TODO: this playbook has not been written yet._
...
 
-_TODO: this playbook has not been written yet._
+This alert fires if Cortex is running without query-scheduler and queries are piling up in the query-frontend queue.
+
+The procedure to investigate it is the same as the one for [`CortexSchedulerQueriesStuck`](#CortexSchedulerQueriesStuck): please see the other playbook for more details.
 
 ### CortexSchedulerQueriesStuck
 
-_TODO: this playbook has not been written yet._
+This alert fires if Cortex is queries are piling up in the query-scheduler.
```
pstibrany, line 417:
```suggestion
This alert fires if queries are piling up in the query-scheduler.
```

cortex-mixin/docs/playbooks.md
```diff
@@ -408,11 +408,31 @@ _TODO: this playbook has not been written yet._
...
+
+How it **works**:
+- A query-frontend API endpoint is called to execute a query
+- The query-frontend enqueues the request to the query-scheduler
+- The query-scheduler is responsible to dispatch enqueued queries to idle querier workers
```
pstibrany, line 422:
```suggestion
- The query-scheduler is responsible for dispatching enqueued queries to idle querier workers
```

cortex-mixin/docs/playbooks.md
```diff
@@ -408,11 +408,31 @@ _TODO: this playbook has not been written yet._
...
+- The querier runs the query, sends the response back directly to the query-frontend and notifies the query-scheduler
```
pstibrany, line 423:
```suggestion
- The querier runs the query, sends the response back directly to the query-frontend and notifies the query-scheduler that it can process another query
```

cortex-mixin/docs/playbooks.md
```diff
@@ -408,11 +408,31 @@ _TODO: this playbook has not been written yet._
...
+- Are queriers in a crash loop (eg. OOMKilled)?
+  - `OOMKilled`: temporarily increase queriers memory request/limit
+  - `panic`: look for the stack trace in the logs and investigate from there
+- Is QPS increased?
+  - Scale up queriers to satisfy the increased workload
+- Is query latency increased?
+  - An increased latency reduces the number of queries we can run / sec: once all workers are busy, new queries will pile up in the queue
+  - Temporarily scale up queriers to try to stop the bleed
+  - Check the `Cortex / Slow Queries` dashboard to see if a specific tenant is running heavy queries
+    - If it's a multi-tenant Cortex cluster and shuffle-sharing is disabled for queriers, you may consider to enable it only for that specific tenant to reduce its blast radius. To enable queriers shuffle-sharding for a single tenant you need to set the `max_queriers_per_tenant` limit override for the specific tenant (the value should be set to the number of queriers assigned to the tenant).
```
pstibrany, line 435:
```suggestion
    - On multi-tenant Cortex cluster with shuffle-sharing for queriers disabled, you may consider to enable it for that specific tenant to reduce its blast radius. To enable queriers shuffle-sharding for a single tenant you need to set the `max_queriers_per_tenant` limit override for the specific tenant (the value should be set to the number of queriers assigned to the tenant).
```

cortex-mixin/docs/playbooks.md
```diff
@@ -408,11 +408,31 @@ _TODO: this playbook has not been written yet._
...
+How to **investigate**:
+- Are queriers in a crash loop (eg. OOMKilled)?
+  - `OOMKilled`: temporarily increase queriers memory request/limit
+  - `panic`: look for the stack trace in the logs and investigate from there
+- Is QPS increased?
+  - Scale up queriers to satisfy the increased workload
+- Is query latency increased?
+  - An increased latency reduces the number of queries we can run / sec: once all workers are busy, new queries will pile up in the queue
+  - Temporarily scale up queriers to try to stop the bleed
+  - Check the `Cortex / Slow Queries` dashboard to see if a specific tenant is running heavy queries
```
pstibrany, line 434: This can perhaps be better checked by looking at `cortex_query_scheduler_queue_length` metric, which is already per-user: `sum by (user) (cortex_query_scheduler_queue_length)`

cortex-mixin/docs/playbooks.md
pstibrany: If user is already configured with querier sharding, and is sending too many slow queries, then nothing from this will help. Arguably, operator could increase the shard size, but that may affect other users too much. Depending on the situation, operator may choose to do nothing, and let Cortex return errors for that given user. (This assumes that only single user is affected).

pstibrany (approved):
Good job on improving playbooks!

pracucci:
Very good suggestions @pstibrany. Could you take another look, please?

PR #342, opened 2021-07-01
Remove CortexQuerierCapacityFull alert

pracucci:
**What this PR does**:
I propose to remove `CortexQuerierCapacityFull` alert. It was added as part of the [initial commit](https://github.com/grafana/cortex-jsonnet/commit/3ff1d4cfcbfa28de1b83c33d42d74749e4c9c97b).

Why removing it? My thoughts:
- The alert fires when a single querier is running queries at full capacity (eg. if configured with 4 workers, all workers run queries for 5 consecutive minutes)
- Having a querier running a full capacity is not an issue per se (if we give a querier 4 workers it means we're willing to use all 4 of them if required) and this looks more alerting on cause then symptom
- We already have alerts on queries stuck in query-frontend/scheduler and we have SLO-based alerts, so I'm not really sure this alert is giving us a real value

N/A

**Checklist**


gouthamve (approved):
LGTM

pracucci:
_I'm not going to merge it immediately. I want to give other people the time to review and think about it._

pstibrany (approved):
Your reasoning makes sense to me. 👍

pracucci:
Going to merge this. I will promptly address any post-merge feedback.

PR #350, opened 2021-07-02
Add playbook for CortexProvisioningTooManyActiveSeries

pracucci:
**What this PR does**:
Added playbook for `CortexProvisioningTooManyActiveSeries` with minor tweaks to the alert.

N/A

**Checklist**

### DIFF
--- a/cortex-mixin/alerts/alerts.libsonnet
+++ b/cortex-mixin/alerts/alerts.libsonnet
@@ -436,19 +436,19 @@
         },
         {
           alert: 'CortexProvisioningTooManyActiveSeries',
-          // 1.5 million active series per ingester max.
+          // We target each ingester to 1.5M in-memory series. This alert fires if the average
+          // number of series / ingester in a Cortex cluster is > 1.6M for 2h (we compact
+          // the TSDB head every 2h).
           expr: |||
             avg by (%s) (cortex_ingester_memory_series) > 1.6e6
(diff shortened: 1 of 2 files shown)


cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -436,19 +436,19 @@
...
         {
           alert: 'CortexProvisioningTooManyActiveSeries',
-          // 1.5 million active series per ingester max.
+          // We target each ingester to 1.5M in-memory series. This alert fires if the average
+          // number of series / ingester in a Cortex cluster is > 1.6M for 2h (we compact
+          // the TSDB head every 2h).
           expr: |||
             avg by (%s) (cortex_ingester_memory_series) > 1.6e6
-              and
-            sum by (%s) (rate(cortex_ingester_received_chunks[1h])) == 0
```
pracucci, line 443: This was just used for the chunks storage chunks transferring.

pstibrany (approved):
This alert feels too similar to `CortexIngesterReachingSeriesLimit` (but that needs instance limits enabled).

pracucci:
> This alert feels too similar to CortexIngesterReachingSeriesLimit (but that needs instance limits enabled).

Exactly! At first I though to remove the alert, but if limit is not enabled then this alert is still useful IMO.

PR #354, opened 2021-07-06
Aded object storage metrics for Ruler and Alertmanager

pracucci:
**What this PR does**:
The Ruler and Alertmanager keeps the config in the object storage, but we don't show the object storage metrics in the Ruler and Alertmanager dashboard. This PR fixes it.

Example:

Screenshot 2021-07-06 at 11 38 53

Screenshot 2021-07-06 at 11 38 40

N/A

**Checklist**

### DIFF
--- a/cortex-mixin/dashboards/alertmanager.libsonnet
+++ b/cortex-mixin/dashboards/alertmanager.libsonnet
@@ -83,5 +83,8 @@ local utils = import 'mixin-utils/utils.libsonnet';
         $.panel('Latency') +
         utils.latencyRecordingRulePanel('cortex_request_duration_seconds', $.jobSelector($._config.job_names.gateway) + [utils.selector.re('route', 'api_v1_alerts|alertmanager')])
       )
+    )
+    .addRows(
+      $.getObjectStoreRows('Alertmanager Configuration Object Store (Alertmanager accesses)', 'alertmanager-storage')
     ),
 }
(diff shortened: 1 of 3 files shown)


cortex-mixin/dashboards/alertmanager.libsonnet
```diff
@@ -83,5 +83,8 @@ local utils = import 'mixin-utils/utils.libsonnet';
         $.panel('Latency') +
         utils.latencyRecordingRulePanel('cortex_request_duration_seconds', $.jobSelector($._config.job_names.gateway) + [utils.selector.re('route', 'api_v1_alerts|alertmanager')])
       )
+    )
+    .addRows(
+      $.getObjectStoreRows('Alertmanager Configuration Object Store (Alertmanager accesses)', 'alertmanager-storage')
```
gotjosh, line 88: out of curiosity, what's the second argument of this function?

gotjosh (approved):
LGTM

cortex-mixin/dashboards/alertmanager.libsonnet
pracucci: The "component" exposing the metrics, which is the name of the internal component using the object store client (used to differentiate all clients when running in single binary mode).

gotjosh: Nice. TIL.

PR #355, opened 2021-07-07
Reduce ingester gRPC max concurrent streams to 10,000

bboreham:
**What this PR does**:
Reduce the limit which governs the number of server calls that gRPC will run in parallel. Beyond this limit, requests will queue in the caller.

While parallelism helps to improve throughput, it is unrealistic that 100,000 operations would be able to proceed at once, considering typical CPU core counts, locking, etc.

Even 10,000 is probably too high, but changing the limit by more than one order of magnitude also seemed rash.

**Checklist**
- [x] `CHANGELOG.md` updated


aknuds1 (approved):

bboreham:
Don't merge yet - want to try it out first.

bboreham:
This change was incorporated in #369.

PR #357, opened 2021-07-14
Clarify data loss related to 'not healthy index found' issue

pracucci:
**What this PR does**:
Following up an offline conversation I had with @jdbaldry, in this PR I'm trying to clarify when there may be a data loss related to block's "not healthy index found" issue.

N/A

**Checklist**


jdbaldry (approved):
Great re-wording, this is much clearer, thank you!

PR #362, opened 2021-07-26
Make CortexIngesterReachingSeriesLimit warning less sensitive

beorn7:
**What this PR does**:

As it turns out, during normal shuffle-sharding operation, the 70%
mark is often exceeded, but not by much. Therefore, this change sets
the new warning mark at 75%. It also increases the `for` duration to
15m as the expected reaction time for warning alerts is usually in the
order of hours, so we can as well wait a bit longer to see if the
problem is transient.

n/a

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 * [ENHANCEMENT] cortex-mixin: Added `alert_excluded_routes` config to exclude specific routes from alerts. #338
 * [ENHANCEMENT] Added `CortexMemcachedRequestErrors` alert. #346
 * [ENHANCEMENT] Ruler dashboard: added "Per route p99 latency" panel in the "Configuration API" row. #353
+* [ENHANCEMENT] Increased the `for` duration of the `CortexIngesterReachingSeriesLimit` warning alert to 3h. #362
 * [BUGFIX] Fixed `CortexIngesterHasNotShippedBlocks` alert false positive in case an ingester instance had ingested samples in the past, then no traffic was received for a long period and then it started receiving samples again. #308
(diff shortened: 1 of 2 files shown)


pracucci (commented):
Thanks Beorn for working on this. I think that what we want from this warning alert is find out ingesters which are constantly above the threshold even after stale series are flushed (which occurs every 2h, when the TSDB head is compacted). Since when the TSDB head is compacted, we flush series with a timestamp between [-3h, -1h] the worst case scenario is that it takes up to 3h to flush stale series. So, I would rather keep the 70% threshold but with a for duration of 3h.

What do you think?

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 * [ENHANCEMENT] cortex-mixin: Added `alert_excluded_routes` config to exclude specific routes from alerts. #338
 * [ENHANCEMENT] Added `CortexMemcachedRequestErrors` alert. #346
 * [ENHANCEMENT] Ruler dashboard: added "Per route p99 latency" panel in the "Configuration API" row. #353
+* [ENHANCEMENT] Tweaked threshould and `for` duration for `CortexIngesterReachingSeriesLimit` warning alert. #362
```
pracucci, line 32: [nit]
```suggestion
* [ENHANCEMENT] Tweaked `for` duration for `CortexIngesterReachingSeriesLimit` warning alert. #362
```

pracucci (approved):

beorn7:
Thanks for your thoughts. Let's make it so!

I have updated the CHANGELOG.md and the commit description accordingly.

beorn7:
Note: I cannot merge this because I'm not authorized. (I guess @pracucci you are. ;-)

PR #365, opened 2021-07-28
Fix scaling dashboard to work on multi-zone ingesters

pracucci:
**What this PR does**:
We have some clusters running Cortex ingesters in multi-zone. Each zone is a StatefulSet whose name matches this pattern `ingester-zone-[a-z]`, so their pod names are like `ingester-zone-a-0` or `ingester-zone-b-1`. All dashboards work correctly except for the scaling dashboard and this PR proposes a fix for that.

Reason why the scaling dashboard doesn't work is that some recording rule computes the expected scaling value summing up all ingesters (eg. `cluster_namespace_deployment_reason:required_replicas:count`) while the actual usage metrics are grouped by deployment/statefulset name, so they're splitted by zone.

I think the desired behaviour is summing up all ingesters, regardless their zone, so in this PR I'm proposing to remove the `-zone-[a-z]` suffix (if found) when we compute the `deployment` name.

A couple of notes:
- I manually tested all updated recording rules and should work
- I had to wrap with another `label_replace()` due to conflicts with regex greediness

N/A

**Checklist**

### DIFF
--- a/cortex-mixin/recording_rules.libsonnet
+++ b/cortex-mixin/recording_rules.libsonnet
@@ -80,10 +80,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
               )
               or
               sum by (cluster, namespace, deployment) (
-                label_replace(
-                  label_replace(kube_statefulset_replicas, "deployment", "$1", "statefulset", "(.*)"),
-                  "deployment", "$1", "deployment", "(.*?)(?:-zone-[a-z])?"
-                )
+                label_replace(kube_statefulset_replicas, "deployment", "$1", "statefulset", "(.*?)(?:-zone-[a-z])?")
               )
             |||,
           },
--- a/cortex-mixin/recording_rules.libsonnet
+++ b/cortex-mixin/recording_rules.libsonnet
@@ -69,12 +69,21 @@ local utils = import 'mixin-utils/utils.libsonnet';
         rules: [
           {
             // Convenience rule to get the number of replicas for both a deployment and a statefulset.
+            // Multi-zone deployments are grouped together removing the "zone-X" suffix.
             record: 'cluster_namespace_deployment:actual_replicas:count',
             expr: |||
-              sum by (cluster, namespace, deployment) (kube_deployment_spec_replicas)
-                or
               sum by (cluster, namespace, deployment) (
-                label_replace(kube_statefulset_replicas, "deployment", "$1", "statefulset", "(.*)")
+                label_replace(
+                  kube_deployment_spec_replicas,
+                  "deployment", "$1", "deployment", "(.*?)(?:-zone-[a-z])?"
+                )
+              )
+              or
+              sum by (cluster, namespace, deployment) (
+                label_replace(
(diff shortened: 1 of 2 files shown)


cortex-mixin/recording_rules.libsonnet
```diff
@@ -69,12 +69,21 @@ local utils = import 'mixin-utils/utils.libsonnet';
...
+            // Multi-zone deployments are grouped together removing the "zone-X" suffix.
             record: 'cluster_namespace_deployment:actual_replicas:count',
             expr: |||
-              sum by (cluster, namespace, deployment) (kube_deployment_spec_replicas)
-                or
               sum by (cluster, namespace, deployment) (
-                label_replace(kube_statefulset_replicas, "deployment", "$1", "statefulset", "(.*)")
+                label_replace(
+                  kube_deployment_spec_replicas,
+                  "deployment", "$1", "deployment", "(.*?)(?:-zone-[a-z])?"
```
tomwilkie, line 78: Why is the first question mark necessary? Shouldn't it be:

```suggestion
                  "deployment", "$1", "deployment", "(.*)(?:-zone-[a-z])?"
```

aknuds1 (approved):

cortex-mixin/recording_rules.libsonnet
pracucci: The first question mark is to make it non-greedy. Since the `(?:-zone-[a-z])?` is optional (ending `?`), if the first `(.*)` is greedy then it always match everything and never removes the zone. Adding `(.*?)` we make the first `.*` non greedy.

cortex-mixin/recording_rules.libsonnet
```diff
@@ -69,12 +69,21 @@ local utils = import 'mixin-utils/utils.libsonnet';
...
+                )
+              )
+              or
+              sum by (cluster, namespace, deployment) (
+                label_replace(
+                  label_replace(kube_statefulset_replicas, "deployment", "$1", "statefulset", "(.*)"),
+                  "deployment", "$1", "deployment", "(.*?)(?:-zone-[a-z])?"
+                )
```
tomwilkie, line 86: The inner label replace is just moving the `statefuleset` label to the deployment label, so could be done with this I believe:

```suggestion
                label_replace(kube_statefulset_replicas, "deployment", "$1", "statefulset", "(.*?)(?:-zone-[a-z])?"),
```

cortex-mixin/recording_rules.libsonnet
tomwilkie: TIL! Can you add a comment to this effect please?

tomwilkie (approved):

cortex-mixin/recording_rules.libsonnet
pracucci: That's right. I've applied the suggested change and manually tested it.

cortex-mixin/recording_rules.libsonnet
pracucci: Sure, done.

PR #366, opened 2021-07-29
Fix rollout dashboard to work with multi-zone deployments

pracucci:
**What this PR does**:
This PR fixes the "Rollout progress" panel of the "Rollout progress" dashboard, to correctly work when a cortex service is deployed across multiples zones.

Notes:
- Follows the same approach took in #365
- I've manually tested the updated queries and should work correctly.

N/A

**Checklist**


stevesg (approved):
LGTM

PR #370, opened 2021-08-16
Clarify the gsutil mv command for moving corrupted blocks

treid314:
Signed-off-by: Tyler Reid <tyler.reid@grafana.com>

**What this PR does**: This PR adds some detail for formulating the gsutil command for moving a corrupted block that would be blocking compaction. This was a point of contention for a recent investigation in which we got an alert but we were unsure of where exactly this data and bucket was kept. 

Fixes #<issue number>

**Checklist**

### DIFF
--- a/cortex-mixin/docs/playbooks.md
+++ b/cortex-mixin/docs/playbooks.md
@@ -380,16 +380,14 @@ When this happen you should:
 2. Ensure the compactor has recovered
 3. Investigate offline the root cause (eg. download the corrupted block and debug it locally)
 
-To rename a block stored on GCS you can use the `gsutil` CLI:
-
+To rename a block stored on GCS you can use the `gsutil` CLI command:
 ```
-# Replace the placeholders:
-# - BUCKET: bucket name
-# - TENANT: tenant ID
-# - BLOCK:  block ID
-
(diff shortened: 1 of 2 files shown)


pstibrany (approved):
👍

cortex-mixin/docs/playbooks.md
```diff
@@ -380,16 +380,14 @@ When this happen you should:
...
-# - BUCKET: bucket name
-# - TENANT: tenant ID
-# - BLOCK:  block ID
-
 gsutil mv gs://BUCKET/TENANT/BLOCK gs://BUCKET/TENANT/corrupted-BLOCK
 ```
+Where:
+- `BUCKET` is the gcs bucket name the compactor is using. The cell's bucket name is specified as the `blocks_storage_bucket_name` in the cell configuration
+- `TENANT` is the tenant id reported in the example error message above as `REDACTED-TENANT`
+- `BLOCK` is the last part of the file path reported as `REDACTED-BLOCK` in the example error message above
```
pracucci, line 390: I think you want to update it the "REDACTED" in the log message above too.

pracucci (approved):
Thanks!

PR #371, opened 2021-08-17
Cortex kv panel improvements

bboreham:
**What this PR does**:

Distributor uses multiple kv stores - for global limits and ha-tracker, as well as reading from the ingester ring - so we need to narrow the panel to just the one it says it is showing, e.g. `kv_name="distributor-hatracker"`.

For consistency, do the same on the ingester panel, although currently ingesters only have one kv store.

Created as Draft because the renamed recording rule will mean that dashboards show no data for latency prior to the change. Do we want to cope with this?  We could add an extra three queries on the latency panels with the old name, but it will get quite ugly in the code. Discuss.

Also, I added a couple of panels with more info on what the KV store is doing.
 - For HA-tracker, show which tenants are changing election.
 - For ingester, show how many are active, leaving, etc.

**Checklist**
- [x] `CHANGELOG.md` updated

### DIFF
--- a/cortex-mixin/dashboards/writes.libsonnet
+++ b/cortex-mixin/dashboards/writes.libsonnet
@@ -110,6 +110,16 @@ local utils = import 'mixin-utils/utils.libsonnet';
         $.panel('Latency') +
         utils.latencyRecordingRulePanel('cortex_kv_request_duration_seconds', $.jobSelector($._config.job_names.distributor) + [utils.selector.eq('kv_name', 'distributor-hatracker')])
       )
+      .addPanel(
+        $.panel('Elected replica changes / min') +
+        $.queryPanel([
+          'max by(exported_cluster, user)(increase(cortex_ha_tracker_elected_replica_changes_total{%s}[1m])) >0' % $.jobMatcher($._config.job_names.distributor),
+        ], [
+          '{{user}}/{{exported_cluster}}',
+        ]) +
+        $.stack + {
+          yaxes: $.yaxes('cpm'),
+        },
       )
     )
     .addRow(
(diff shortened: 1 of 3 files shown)


cortex-mixin/dashboards/writes.libsonnet
```diff
@@ -104,11 +104,22 @@ local utils = import 'mixin-utils/utils.libsonnet';
...
       )
       .addPanel(
         $.panel('Latency') +
-        utils.latencyRecordingRulePanel('cortex_kv_request_duration_seconds', $.jobSelector($._config.job_names.distributor))
+        utils.latencyRecordingRulePanel('cortex_kv_request_duration_seconds', $.jobSelector($._config.job_names.distributor) + [utils.selector.eq('kv_name', 'distributor-hatracker')])
+      )
+      .addPanel(
+        $.panel('Elected replica changes / min') +
+        $.queryPanel([
+          'max by(exported_cluster, user)(increase(cortex_ha_tracker_elected_replica_changes_total{%s}[1m])) >0' % $.jobMatcher($._config.job_names.distributor),
```
pracucci, line 116: This can be a pretty high cardinality query. For example, I just tested in a cluster with few thousands of tenants and running it over "Last 12h" took 25s on cold caches. What if we add a recording rule for that? I would expect the `> 0` narrows down a lot the cardinality (other than squashing it by a "distributors replicas" number factor).

pracucci (commented):
> Created as Draft because the renamed recording rule will mean that dashboards show no data for latency prior to the change. Do we want to cope with this? We could add an extra three queries on the latency panels with the old name, but it will get quite ugly in the code. Discuss.

If it was the read/write latency, I would have said definitely yes. In this case it's the KV latency so we can probably accept the fact we'll "loose the history" when watching this dashboard in the past (or over la large period), so to me it's OK.

tleopold:

PR #379, opened 2021-08-25
fix(rules): upstream recording rule switched to sum_irate

Duologic:
ref: https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/619

**What this PR does**:
Upstream recording rule switched to sum_irate, this was reflected in the recording rule name.

**Checklist**

### DIFF
--- a/cortex-mixin/recording_rules.libsonnet
+++ b/cortex-mixin/recording_rules.libsonnet
@@ -213,7 +213,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
               sum by (cluster, namespace, deployment) (
                 label_replace(
                   label_replace(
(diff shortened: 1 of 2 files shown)


cortex-mixin/recording_rules.libsonnet
```diff
@@ -213,7 +213,7 @@ local utils = import 'mixin-utils/utils.libsonnet';
               sum by (cluster, namespace, deployment) (
                 label_replace(
                   label_replace(
-                    node_namespace_pod_container:container_cpu_usage_seconds_total:sum_rate,
+                    node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate,
```
Duologic, line 216: I'm not sure if we should reflect that in the name of this recording rule too.

gouthamve (approved):
LGTM

PR #385, opened 2021-09-02
Fixes to initial state sync panels on alertmanager dashboard.

stevesg:
**What this PR does**:

1) Change minimal interval to 1m for sync duration and fetch state panels.

    This is in order to show infrequent events at smaller time windows.

2) Change syncs/sec panel to reflect absolute value of metric not rate.

    The initial sync only occurs once per-tenant so the counter value is
    essentially 0 or 1. Due to how per-tenant metrics are aggregated, the
    external facing metric really acts more like a gauge reflecting the number
    of tenants which achieved each outcome.

    Also, stack this panel as it becomes easier to visually see when the initial
    syncs have completed for all tenants (e.g. during a rollout).

**Checklist**
- [X] ~~`CHANGELOG.md` updated - the order of entries should be `[CHANGE]`, `[FEATURE]`, `[ENHANCEMENT]`, `[BUGFIX]`~~

### DIFF
--- a/cortex-mixin/dashboards/alertmanager.libsonnet
+++ b/cortex-mixin/dashboards/alertmanager.libsonnet
@@ -152,15 +152,23 @@ local utils = import 'mixin-utils/utils.libsonnet';
     .addRow(
       $.row('Sharding Initial State Sync')
       .addPanel(
-        $.panel('Initial syncs/sec') +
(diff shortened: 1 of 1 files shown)


gotjosh (approved):
LGTM

PR #386, opened 2021-09-02
Add overrides to compactor

gouthamve:


pstibrany (approved):
I think this deserves changelog entry too.

PR #387, opened 2021-09-02
Add recording rules for Alertmanager dashboard,

stevesg:
**What this PR does**:
With large numbers of tenants the queries for these panels can become quite
slow as the metrics exposed are per-tenant.

**Checklist**

### DIFF
--- a/cortex-mixin/dashboards/alertmanager.libsonnet
+++ b/cortex-mixin/dashboards/alertmanager.libsonnet
@@ -11,11 +11,11 @@ local utils = import 'mixin-utils/utils.libsonnet';
        })
       .addPanel(
         $.panel('Total Alerts') +
-        $.statPanel('sum(cortex_alertmanager_alerts{%s})' % $.jobMatcher('alertmanager'), format='short')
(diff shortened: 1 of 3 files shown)


cortex-mixin/dashboards/alertmanager.libsonnet
```diff
@@ -11,11 +11,11 @@ local utils = import 'mixin-utils/utils.libsonnet';
        })
       .addPanel(
         $.panel('Total Alerts') +
-        $.statPanel('sum(cortex_alertmanager_alerts{%s})' % $.jobMatcher('alertmanager'), format='short')
+        $.statPanel('sum(cluster_job_%s:cortex_alertmanager_alerts:sum{%s})' % [$._config.per_instance_label, $.jobMatcher('alertmanager')], format='short')
```
gotjosh, line 14: If the recording rule is a sum, do we need to sum here again?

gotjosh (approved):
LGTM - but have one question.

cortex-mixin/dashboards/alertmanager.libsonnet
stevesg: Good question, I also wondered about this, but in general I decided to leave all the sum()s as-is because the dashboard allows selecting multiple cluster/namespaces.

gotjosh (approved):
LGTM

pracucci (approved):
LGTM, thanks!

PR #389, opened 2021-09-07
Make the overrides metric name configurable.

gouthamve:
We (Grafana Labs) are about to put in a new system to control and export
data about limits and we'll need to use a different name. This shouldn't
affect our OSS users.


bboreham (approved):
seems legit

PR #391, opened 2021-09-13
Split limits config into a variable we can reuse

gouthamve:


cortex/distributor.libsonnet
```diff
@@ -30,8 +31,6 @@
       'server.grpc.keepalive.max-connection-idle': '1m',
 
       'distributor.ingestion-rate-limit-strategy': 'global',
```
pracucci, line 33: [nit] I would move this too.

pracucci (approved):

PR #398, opened 2021-09-20
Removed chunks storage query sharding config support

pracucci:
**What this PR does**:
Chunks storage query sharding never hit production. To remove some misunderstanding, I would suggest to remove its support from this mixin config.

N/A

**Checklist**


cyriltovena (approved):
LGTM

PR #403, opened 2021-10-12
Use `$._config.job_names.gateway` in resources dashboards.

stevesg:
This fixes panels where `cortex-gw` was hardcoded.


stevesg:
Tested now - will merge.

PR #405, opened 2021-10-13
Add CortexRolloutStuck alert

pracucci:
**What this PR does**:
In this PR I propose to add the `CortexRolloutStuck` which fires if a Cortex StatefulSet or Deployment rollout is stuck. I've manually tried the queries and they should work as expected. The alert is a warning to get some confidence with it, but final goal would be to run it as critical.

N/A

**Checklist**

### DIFF
--- a/cortex-mixin/docs/playbooks.md
+++ b/cortex-mixin/docs/playbooks.md
@@ -231,16 +231,6 @@ How to **investigate**:
 
 _If the alert `CortexIngesterTSDBHeadCompactionFailed` fired as well, then give priority to it because that could be the cause._
 
-### CortexRolloutStuck
-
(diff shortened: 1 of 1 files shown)


cortex-mixin/alerts/alerts.libsonnet
```diff
@@ -412,6 +412,67 @@
...
+      rules: [
+        {
+          alert: 'CortexRolloutStuck',
+          expr: |||
+            (
+              max without (revision) (
+                kube_statefulset_status_current_revision
+                  unless
+                kube_statefulset_status_update_revision
+              )
```
simonswine, line 426: I am not too sure why stateful sets need that revision check, but I guess it's also upstream: https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/6c72589035f4f49674a56cf97a3ec1a02f14671a/alerts/apps_alerts.libsonnet#L128

So it should be ok :slightly_smiling_face:

pracucci: Yeah, I learned it from there.

PR #411, opened 2021-10-18
use alertmanager jobname for alertmanager dashboard panels

aallawala:
**What this PR does**:
This PR adjusts the alertmanager dashboard to use the _config.job_names.alertmanager instead of the hardcoded 'alertmanager' value.
Fixes #<issue number>

**Checklist**

### DIFF
--- a/cortex-mixin/config.libsonnet
+++ b/cortex-mixin/config.libsonnet
@@ -37,7 +37,7 @@
       store_gateway: '(store-gateway|cortex$)',
       gateway: '(gateway|cortex-gw|cortex-gw-internal)',
       compactor: 'compactor.*',  // Match also custom compactor deployments.
-      alertmanager: '(alertmanager.*|cortex$)',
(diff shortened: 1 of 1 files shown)


cortex-mixin/config.libsonnet
```diff
@@ -37,6 +37,7 @@
       store_gateway: '(store-gateway|cortex$)',
       gateway: '(gateway|cortex-gw|cortex-gw-internal)',
       compactor: 'compactor.*',  // Match also custom compactor deployments.
+      alertmanager: '(alertmanager.*|cortex$)',
```
pracucci, line 40: I would just use "alertmanager" as default. The alertmanager is not running as part of the single binary (at least not by default) so I would avoid having `|cortex`. I would also avoid `alertmanager.*` by default since it could match other unrelated alertmanager installations (eg. Prometheus's vanilla one).

```suggestion
      alertmanager: 'alertmanager',
```

PR #412, opened 2021-10-18
Span the annotation.message in alerts as YAML multiline strings.

aallawala:
**What this PR does**:
This PR attempts to fix the issue where the annotation message is split across two lines instead of one. This causes an issue for our alertmanager config template to not render correctly since it splits up the `printf` template across the two lines.

Before:
```yaml
groups:
- name: cortex_alerts
  rules:
  - alert: CortexIngesterUnhealthy
    annotations:
      message: Cortex cluster {{ $labels.cluster }}/{{ $labels.namespace }} has {{
        printf "%f" $value }} unhealthy ingester(s).
    expr: |
      min by (cluster, namespace) (cortex_ring_members{state="Unhealthy", name="ingester"}) > 0
    for: 15m
    labels:
      severity: critical
```

After
```yaml
groups:
- name: cortex_alerts
  rules:
  - alert: CortexIngesterUnhealthy
    annotations:
      message: |
        Cortex cluster {{ $labels.cluster }}/{{ $labels.namespace }} has {{ printf "%f" $value }} unhealthy ingester(s).
    expr: |
      min by (cluster, namespace) (cortex_ring_members{state="Unhealthy", name="ingester"}) > 0
    for: 15m
    labels:
      severity: critical
```
Fixes #<issue number>

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,7 +77,7 @@
 * [BUGFIX] Upstream recording rule `node_namespace_pod_container:container_cpu_usage_seconds_total:sum_irate` renamed. #379
 * [BUGFIX] Treat `compactor_blocks_retention_period` type as string rather than int.#395
 * [BUGFIX] Fixed writes/reads/alertmanager resources dashboards to use `$._config.job_names.gateway`. #403
-* [BUGFIX] Span the annotation.message field for `CortexIngesterUnhealthy` alert in one line. #412
+* [BUGFIX] Span the annotation.message in alerts as YAML multiline strings. #412
 
 ## 1.9.0 / 2021-05-18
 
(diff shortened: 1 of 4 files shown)


pracucci (approved):
Changes LGTM, but why does this even fix your issue? Could you clarify, please?

aallawala:
> Changes LGTM, but why does this even fix your issue? Could you clarify, please?

Thank you @pracucci for taking a look. 

Spanning the template across multiple lines affects us since we utilize a separate Go templating layer on top of the Go templating layer expected by the alertmanager. Our templates in code without this change look something like this:

```yaml
groups:
- name: cortex_alerts
  rules:
  - alert: CortexIngesterUnhealthy
    annotations:
      message: Cortex cluster {{ `{{ "{{ $labels.cluster }}" }}` }}/{{ `{{ "{{ $labels.namespace }}" }}` }} has {{ `{{ "{{
        printf \"%.*f\" $value }}" }}` }} unhealthy ingester(s).
    expr: |
      min by (cluster, namespace) (cortex_ring_members{state="Unhealthy", name="ingester"}) > 0
    for: 15m
    labels:
      severity: critical
```

The line delimit prior to the `printf` falls between a {{ }} that causes the Go templating to fail.

pracucci:
> > I'm still not sure why this change outputs the message on a single line instead only cutting it. That was my question. Could you elaborate on it, please?
> 
> From what I understand from the jsonnet [spec](https://jsonnet.org/ref/spec.html), wrapping the message with a `|||` in the jsonnet template preserves the line endings whereas the single-quoted `'` and double-quoted `"` strings are allowed to span multiple lines in the output of the jsonnet.

I didn't know it, but now it makes perfect sense.

PR #415, opened 2021-11-09
cortex: Make ruler object storage support generic

aknuds1:
**What this PR does**:
Make ruler object storage support generic, which in turn fixes EKS/S3 support.

**Checklist**

### DIFF
--- a/cortex/config.libsonnet
+++ b/cortex/config.libsonnet
@@ -244,8 +244,7 @@
 
     ruler_enabled: false,
     ruler_client_type: error 'you must specify a storage backend type for the ruler (azure, gcs, s3, local)',
-    ruler_s3_bucket_name: error 'you must specify the ruler S3 bucket name',
-    ruler_gcs_bucket_name: error 'must specify a GCS bucket name',
+    ruler_storage_bucket_name: error 'must specify the ruler storage bucket name',
 
     rulerClientConfig:
       {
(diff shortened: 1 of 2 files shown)


cortex/config.libsonnet
```diff
@@ -244,20 +244,23 @@
...
-          'ruler-storage.gcs.bucket-name': $._config.ruler_gcs_bucket_name,
+          'ruler-storage.gcs.bucket-name': $._config.ruler_storage_bucket_name,
         },
         s3: {
           'ruler-storage.s3.region': $._config.aws_region,
-          'ruler-storage.s3.bucket-name': $._config.ruler_s3_bucket_name,
+          'ruler-storage.s3.bucket-name': $._config.ruler_storage_bucket_name,
+        },
+        azure: {
+          // TODO: Is this the correct flag??
```
pracucci, line 262: Config is a bit different for azure:

```
azure:
  # Azure storage account name
  # CLI flag: -ruler-storage.azure.account-name
  [account_name: <string> | default = ""]

  # Azure storage account key
  # CLI flag: -ruler-storage.azure.account-key
  [account_key: <string> | default = ""]

  # Azure storage container name
  # CLI flag: -ruler-storage.azure.container-name
  [container_name: <string> | default = ""]

  # Azure storage endpoint suffix without schema. The account name will be
  # prefixed to this value to create the FQDN
  # CLI flag: -ruler-storage.azure.endpoint-suffix
  [endpoint_suffix: <string> | default = ""]

  # Number of retries for recoverable errors
  # CLI flag: -ruler-storage.azure.max-retries
  [max_retries: <int> | default = 20]
```

aknuds1: @pracucci at the moment, this flag is used for Azure, so removing it will cause a change in Flux manifests, but I guess it should have no effect, based on the documentation you pasted.

I can safely remove it in this PR then? The Azure flags you pasted above are not configured in cortex-jsonnet at the moment, but instead directly in our Azure cell config.

aknuds1: @pracucci I removed `ruler-storage.gcs.bucket-name` for Azure, as you clarified not relevant.

pracucci (approved):

PR #427, opened 2022-01-24
Update dashboards to support Helm chart deployment

### LINKED ISSUE #361: dealing with mostly broken Grafana dashboards
I built dashboards following the [mixtool instructions](https://github.com/grafana/cortex-jsonnet#monitoring-for-cortex), got the dashboards uploaded in Grafana and I found myself lots in a sea of "No data". Most dashdoards fall in that category: reads, writes, compactor resources, queries etc, only the object store shows half the graphs working.

Digging one of the queries tells me the issues are with how the queries were implemented:

`sum(cluster_namespace_job:cortex_distributor_received_samples:rate5m{cluster=~"$cluster", **job=~"($namespace)/(distributor|cortex$)"}**)`

The metric above has different labels out of Prometheus, there is no way it would work out of the box:

`cluster_namespace_job:cortex_distributor_received_samples:rate5m{**job="cortex-distributor"**, namespace="cortex"}`

More or less, same is true for all other "No data" charts, it looks like each needs to be changed to make it work.

I instaled Cortex from the latest version of [your Helm chart](https://github.com/cortexproject/cortex-helm-chart) (ver 0.6.0 at the time) and it does get the latest cortex image deployes, so the metric labels should be up to date.

Before I take on the monumental task of fixing these charts I wanted to ask, is there something I missed ? My expectation was, by havind the latest cortex deployed from the latest Helm chart, that most if not all of the dashboards provided would come with queries matching the metrics exposed.

danfromtitan:
**What this PR does**:

Update cortex-mixin to support Helm chart deployment and produce usable dashboards. List of issues fixed by this change:
- support disabling cortex-gw, query scheduler and ruler panels
- generalize instance names to also match Helm chart deployments
- remove cluster from namespace matcher to cope with metrics outside Cortex service that don't have a cluster label
- fix disk panels to display correct written byte metrics
- support custom Prometheus http prefix
- display read/write metrics from Cortex distributor/querier instead of cortex-gw
- rearrange panels in rollout dashboard to make better sense of the metrics shown

Caveat: 
- I tested dashboards with Helm chart deployments but since that uses opensource Cortex image I could not test the dashboards when cortex-gw component is enabled.  

Fixes #361

**Checklist**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## master / unreleased
 
+* [ENHANCEMENT] Update cortex-mixin to support Cortex deployment from the Helm chart. #361
+
 ## 1.11.0 / 2021-12-30
 
 * [CHANGE] Store gateway: set `-blocks-storage.bucket-store.index-cache.memcached.max-get-multi-concurrency`,
--- a/cortex-mixin/config.libsonnet
+++ b/cortex-mixin/config.libsonnet
@@ -7,6 +7,11 @@
     // May contain 'chunks', 'blocks' or both.
     // Enables chunks- or blocks- specific panels and dashboards.
     storage_engine: ['blocks'],
+    
+    // Disable unused panels depending whether a component was installed or not
+    cortex_gw_enabled: false,
+    query_scheduler_enabled: false,
+    ruler_enabled: false,
 
     // For chunks backend, switch for chunk index type.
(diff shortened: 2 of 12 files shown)


jmcarp:
Do you all have capacity to review this @pracucci? Would be useful for my team as well.

tleopold:

